PR #6, opened 2020-04-03
Metric for schedulers

inventormc:


tune_sklearn/tune_search.py
```diff
@@ -292,6 +292,7 @@ def __init__(
     ):
         self.estimator = estimator
         self.scheduler = scheduler
+        scheduler.metric = "average_test_score"
```
anthonyhsyu, line 295: Needs to handle if the user does not pass in a scheduler

tune_sklearn/tune_search.py
```diff
@@ -292,7 +292,8 @@ def __init__(
     ):
         self.estimator = estimator
         self.scheduler = scheduler
-        scheduler.metric = "average_test_score"
+        if self.scheduler != None:
```
anthonyhsyu, line 295:
```suggestion
        if self.scheduler is not None:
```

PR #11, opened 2020-04-10
Support string schedulers

inventormc:


tune_sklearn/tune_search.py
```diff
@@ -380,27 +385,49 @@ def __init__(
...
+                        self.scheduler = HyperBandForBOHB(
+                            metric="average_test_score")
+                    elif scheduler == "MedianStoppingRule":
+                        self.scheduler = MedianStoppingRule(
+                            metric="average_test_score")
+                else:
+                    raise ValueError("{} is not a defined scheduler. "
+                                     "Check the list of available schedulers."
+                                     .format(scheduler))
+            else:
```
anthonyhsyu, line 415: Should check whether the scheduler is an instance of Ray Tune's `TrialScheduler` object, and error if the user passes in something besides a string or a `TrialScheduler`

PR #14, opened 2020-04-17
Add examples

anthonyhsyu:
LightGBM example + pipeline example

Could be merged, and we can open a new PR for #17 which caused us problems before that had nothing to do with pipelines.


inventormc:
It looks like the `is_pipeline` isn't being used in the code, so we can remove it.

PR #21, opened 2020-05-01
Bayesian Optimization

inventormc:
Add functionality for Bayesian optimization for searching over distributions.

### DIFF
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -885,13 +885,12 @@ class TuneSearchCV(TuneBaseSearchCV):
             if search_optimization == "random":
                 if not (isinstance(dist, list) or hasattr(dist, "rvs")):
                     raise ValueError(
-                        "distribution must be list or scipy "
+                        "distribution must be a list or scipy "
                         "distribution when using randomized search")
             else:
                 if not isinstance(dist, tuple):
-                    raise ValueError(
-                        "distribution must be tuple when using bayesian search"
-                    )
+                    raise ValueError("distribution must be a tuple when using "
+                                     "bayesian search")
 
         super(TuneSearchCV, self).__init__(
             estimator=estimator,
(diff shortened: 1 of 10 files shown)


tune_sklearn/tune_search.py
```diff
@@ -888,6 +918,9 @@ def _fill_config_hyperparam(self, config):
                 configuration for `tune.run`.
 
         """
+        if self.search_optimization == "bayesian":
+            return
```
anthonyhsyu, line 922: Not sure if this is right. I know you're passing this into `BayesOptSearch`'s `space` parameter which seems correct, but I'm not sure if leaving this out of the `config` dictionary would ruin anything. Just double check to make sure this is good!

tune_sklearn/tune_search.py
```diff
@@ -1142,27 +1182,18 @@ def _tune_run(self, config, resources_per_trial):
...
-                scheduler=self.scheduler,
-                reuse_actors=True,
-                verbose=self.verbose,
-                stop={"training_iteration": self.early_stopping_max_epochs},
-                config=config,
-                checkpoint_at_end=True,
-                resources_per_trial=resources_per_trial,
-            )
+
+        analysis = tune.run(
```
anthonyhsyu, line 1188: Nice refactor lol

tune_sklearn/tune_search.py
```diff
@@ -840,25 +846,47 @@ class TuneRandomizedSearchCV(TuneBaseSearchCV):
...
+            raise ValueError("Search optimization must be random or bayesian")
+        if search_optimization == "bayesian" and random_state is not None:
+            warnings.warn(
+                "random state is ignored when using Bayesian optimization")
+
+        for dist in param_distributions.values():
+            if search_optimization == "random":
+                if not (isinstance(dist, list) or hasattr(dist, "rvs")):
+                    raise ValueError(
+                        "distribution must be list or scipy "
```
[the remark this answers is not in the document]
inventormc: example has been added (bayesian_sgd.py)

tune_sklearn/tune_search.py
inventormc: This seems to be okay based on tune's example; the only parameter passed into config is iterations which doesn't look like a hyperparameter being tuned.

https://github.com/ray-project/ray/blob/master/python/ray/tune/examples/bayesopt_example.py

PR #23, opened 2020-05-08
Early stopping

inventormc:
Added `early_stopping` parameter back in and add default ASHA scheduler.

### DIFF
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -28,7 +28,7 @@ from ray import tune
 from ray.tune import Trainable
 from ray.tune.schedulers import (
     PopulationBasedTraining, AsyncHyperBandScheduler, HyperBandScheduler,
-    HyperBandForBOHB, MedianStoppingRule, TrialScheduler)
+    HyperBandForBOHB, MedianStoppingRule, TrialScheduler, ASHAScheduler)
 from ray.tune.suggest.bayesopt import BayesOptSearch
 import numpy as np
 from numpy.ma import MaskedArray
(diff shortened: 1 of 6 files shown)


tune_sklearn/tune_search.py
```diff
@@ -406,17 +411,21 @@ def __init__(
...
+                self.scheduler.metric = "average_test_score"
+            elif scheduler is None:
+                self.scheduler = ASHAScheduler(metric="average_test_score")
             else:
                 raise TypeError("Scheduler must be a str or tune scheduler")
-        else:
+        elif self.early_stopping:
             warnings.warn("Unable to do early stopping because "
                           "estimator does not have `partial_fit`")
             self.early_stopping_max_epochs = 1
```
anthonyhsyu, line 431: Add `else` case to set `self.scheduler` to something if none of these if cases are satisfied

tune_sklearn/tune_search.py
```diff
@@ -425,11 +426,14 @@ def __init__(
                 self.scheduler = ASHAScheduler(metric="average_test_score")
             else:
                 raise TypeError("Scheduler must be a str or tune scheduler")
-        elif self.early_stopping:
-            warnings.warn("Unable to do early stopping because "
-                          "estimator does not have `partial_fit`")
+        else:
+            warnings.warn("Early stopping is not enabled. "
+                          "This may be because the estimator "
+                          "does not have `partial_fit`"
```
anthonyhsyu, line 432:
```suggestion
                          "does not have `partial_fit`")
```

tune_sklearn/tune_search.py
```diff
@@ -425,11 +426,14 @@ def __init__(
...
+
             self.early_stopping_max_epochs = 1
```
anthonyhsyu, line 434: `early_stopping_max_epochs` to `max_iter`

PR #25, opened 2020-05-17
List of dictionaries

inventormc:
Add support for grid search over a list of grids.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ script:
   - python3 random_forest.py
   - python3 sgd.py
   - python3 torch_nn.py
+  - python3 xgbclassifier.py
 
 cache:
   directories:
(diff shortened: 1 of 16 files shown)


inventormc:
This should work for grid search (still needs to be tested), but this solution won't work for `TuneSearchCV` because the `search_alg` parameter may need to be set to Bayesian optimization. Does it even make sense to implement this feature for searching over distributions? Sklearn's `ParameterSampler` generator doesn't expect a list while its `ParameterGrid` generator does.

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.ParameterGrid.html#sklearn.model_selection.ParameterGrid

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.ParameterSampler.html#sklearn.model_selection.ParameterSampler

anthonyhsyu:
Yes, ideally we would need this supported for Bayesian as well, as one of the ways it can have it's `search_space` specified is by a list:
https://scikit-optimize.github.io/stable/modules/generated/skopt.BayesSearchCV.html

PR #29, opened 2020-06-05
Split code

inventormc:
tune_search.py is getting really long


tune_sklearn/tune_search.py
```diff
@@ -9,713 +9,16 @@
     -- Anthony Yu and Michael Chau
 """
```
anthonyhsyu, line 10: Let's get this above documentation in the other files too

PR #30, opened 2020-06-23
parallelism for no early stopping

anthonyhsyu:
small but important change when early stopping is false


inventormc:
Changes look good to me but checks are failing on the keras example. There's an issue with the parameters being passed in (might be because of a new version).

PR #32, opened 2020-06-28
"scheduler" removed for simpler interface

anthonyhsyu:


examples/random_forest.py
```diff
@@ -25,7 +25,6 @@
     clf,
     param_distributions,
     n_iter=3,
-    max_iters=10,
```
inventormc, line 28: Why did we remove `max_iters` here? Not that it makes a huge difference in terms of showing the example but jw.

tests/test_gridsearch.py
```diff
@@ -114,13 +113,12 @@ def test_grid_search_no_score(self):
         # in wrong results. This only happens with threads, not processes/sync.
         # For now, we'll fit using the sync scheduler.
         grid_search = TuneGridSearchCV(
-            clf, {"C": Cs}, scoring="accuracy", scheduler=MedianStoppingRule())
```
inventormc, line 117: Why didn't we replace scheduler with early_stopping here?

tests/test_gridsearch.py
```diff
@@ -114,13 +113,12 @@ def test_grid_search_no_score(self):
...
+            clf, {"C": Cs}, scoring="accuracy")
         grid_search.fit(X, y)
 
         grid_search_no_score = TuneGridSearchCV(
             clf_no_score, {"C": Cs},
-            scoring="accuracy",
-            scheduler=MedianStoppingRule())
```
inventormc, line 123: Why didn't we replace scheduler with early_stopping here?

tune_sklearn/_trainable.py
```diff
@@ -42,13 +42,12 @@ def _setup(self, config):
...
         self.fit_params = config.pop("fit_params")
         self.scoring = config.pop("scoring")
-        self.early_stopping = config.pop("early_stopping")
         self.max_iters = config.pop("max_iters")
         self.cv = config.pop("cv")
         self.return_train_score = config.pop("return_train_score")
         self.estimator_config = config
 
-        if self.early_stopping:
+        if self.early_stopping is not None:
```
inventormc, line 50: Should we also have the same conditional in the _train method (line 79) `if self.early_stopping is not None:`. Just so it's more clear and matches with this part.

tune_sklearn/tune_basesearch.py
```diff
@@ -194,51 +194,52 @@ def _check_is_fitted(self, method_name):
...
-            elif isinstance(scheduler, TrialScheduler):
-                self.scheduler = scheduler
-                self.scheduler.metric = "average_test_score"
-            elif scheduler is None:
-                self.scheduler = ASHAScheduler(metric="average_test_score")
+                                     .format(early_stopping))
+            elif isinstance(early_stopping, TrialScheduler):
+                self.early_stopping = early_stopping
+                self.early_stopping.metric = "average_test_score"
+            elif early_stopping is None:
```
inventormc, line 238: This case can never be hit because `early_stopping` cannot be None here. I think this case is covered by the outer else clause.

tune_sklearn/_trainable.py
inventormc: Actually there's a few places where both versions of this check is used in all the other files (tune_basesearch.py, tune_gridsearch.py, tune_search.py). Doesn't `if self.early_stopping:` do the same thing as `if self.early_stopping is not None:`? Either way it would be nice to have it consistent.

examples/random_forest.py
anthonyhsyu: We didn't activate `early_stopping` in this example, so setting `max_iters` doesn't do anything. I removed it to prevent confusion of how this parameter is used

tests/test_gridsearch.py
anthonyhsyu: We could, but in spirit of performing the original tests I decided to remove it.

tests/test_gridsearch.py
anthonyhsyu: We could, but in spirit of performing the original tests I decided to remove it.

tune_sklearn/tune_basesearch.py
anthonyhsyu: Good catch thanks

tune_sklearn/_trainable.py
anthonyhsyu: Yep, missed some spots and will fix

PR #34, opened 2020-06-28
Gpu/cpu support

inventormc:
Added support for user to specify number of gpu and cpu. This removed the need for `n_jobs`; need to confirm that these new changes behave correctly/don't negatively impact performance.


richardliaw:
Hey, I think instead we should use `num_jobs`, but also enable `use_gpu`.

PR #36, opened 2020-07-04
Documentation - proper rendering.

richardliaw:
This PR consists of two main changes to make the docstrings renderable on docs.ray.io - 

1. remove :ref: to Scikit-learn specific concepts
2. Proper white-spacing/indentations (readthedocs is sensitive to newlines, so I removed as many of them as I could.)

### DIFF
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -34,8 +34,7 @@ class TuneBaseSearchCV(BaseEstimator):
 
     defined_schedulers = [
         "PopulationBasedTraining", "AsyncHyperBandScheduler",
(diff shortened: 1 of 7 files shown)


tune_sklearn/tune_basesearch.py
```diff
@@ -220,9 +221,6 @@ def __init__(self,
                     elif early_stopping == "HyperBandScheduler":
                         self.early_stopping = HyperBandScheduler(
                             metric="average_test_score")
-                    elif early_stopping == "HyperBandForBOHB":
-                        self.early_stopping = HyperBandForBOHB(
-                            metric="average_test_score")
```
richardliaw, line 225: I don't think HyperBandForBOHB actually works here

PR #37, opened 2020-07-11
Remove ray license

richardliaw:
This removes licenses irrelevant to tune-sklearn. Did we reference sklearn in our impl? perhaps we should update this then.


anthonyhsyu:
What do you mean by referencing sklearn in our implementation? There are some functions we use from sklearn, if that's what you mean.

richardliaw:
Yes, that we share code from their repo. We should be sure to list their license appropriately.

PR #40, opened 2020-07-13
Early stop bug fix

inventormc:
`self.early_stopping` was not set properly when using `early_stopping=True`

### DIFF
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -209,8 +209,8 @@ class TuneBaseSearchCV(BaseEstimator):
         if early_stopping and self._can_early_stop():
             self.max_iters = max_iters
             if early_stopping is True:
-                self.early_stopping = "AsyncHyperBandScheduler"
-            elif isinstance(early_stopping, str):
+                early_stopping = "AsyncHyperBandScheduler"
+            if isinstance(early_stopping, str):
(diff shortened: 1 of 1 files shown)


tune_sklearn/tune_basesearch.py
```diff
@@ -209,7 +209,7 @@ def __init__(self,
         if early_stopping and self._can_early_stop():
             self.max_iters = max_iters
             if early_stopping is True:
-                early_stopping = "AsyncHyperBandScheduler"
+                self.early_stopping = "AsyncHyperBandScheduler"
             elif isinstance(early_stopping, str):
```
richardliaw, line 213: I think this doesn't work; 

the idea here was to make early_stopping = "AsyncHyperBandScheduer" so it is checked in L214.

richardliaw: otherwise, self.early_stopping is never actually set into a scheduler object

richardliaw, line 213:
```suggestion
                # Override the early_stopping variable so
                # that it is resolved appropriately in
                # the next block
                early_stopping = "AsyncHyperBandScheduler"

            # Resolve the early stopping object
            if isinstance(early_stopping, str):
```

inventormc: Whoops yeah you're right. Changed to address this.

PR #41, opened 2020-07-13
Add ray test to travis

richardliaw:


.travis.yml
```diff
@@ -44,6 +44,7 @@ script:
   - cd tests
   - python3 -m unittest test_randomizedsearch.py
   - python3 -m unittest test_gridsearch.py
+  - cd ~/ && git clone https://github.com/ray-project/ray && python3 ray/doc/source/tune/_tutorials/tune-sklearn.py
```
inventormc, line 47: I think this might mess with the next line where we run stuff in the examples folder because `../examples` will be the wrong path. We could just move this line down to after the examples stuff.

.travis.yml
```diff
@@ -44,6 +44,7 @@ script:
...
   - cd ../examples
   - for f in *.py; do python3 "$f"; done
```
richardliaw, line 49:
```suggestion
  - cd ../examples
  - for f in *.py; do python3 "$f"; done
  - cd ~/ && git clone https://github.com/ray-project/ray && python3 ray/doc/source/tune/_tutorials/tune-sklearn.py
```

inventormc:
The import from the test uses `ray.tune.sklearn` which doesn't work since travis uses the stable build of ray. We would need the nightly build for this test to work.

anthonyhsyu:
Is there a timeline in which this would be merged into the stable build of ray?

richardliaw:
In a couple weeks, but I just pushed a workaround.

.travis.yml
```diff
@@ -5,14 +5,16 @@ matrix:
   include:
     - os: linux
       python: 3.6
-      env: PYTHON=3.6
+      env: PYTHON=3.6 OS=LINUX
+      before_script: pip install -U https://s3-us-west-2.amazonaws.com/ray-wheels/latest/ray-0.9.0.dev0-cp36-cp36m-manylinux1_x86_64.whl
     - os: osx
       osx_image: xcode9.4
       language: generic
       env: PYTHON=3.6 PYTHONWARNINGS=ignore OS=MAC
```
inventormc, line 13: Just curious, why do the mac tests not need the nightly wheels?

richardliaw: ah because it looks like the nightly ray wheels aren't supported on this version of osx.

PR #43, opened 2020-07-16
Parallelize partial_fit

inventormc:
Use joblib to parallelize the folds of partial_fit/early stop,

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -11,7 +11,7 @@ import os
 from pickle import PicklingError
 import ray.cloudpickle as cpickle
 import warnings
-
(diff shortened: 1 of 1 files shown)


inventormc:
> Did you observe any improvements in runtimes with this change?

Not with initial tests. Could be because of the `njobs=1`, but other values makes ray stall.

anthonyhsyu:
Should we close this PR for now? If it doesn't do anything maybe we can leave it alone

inventormc:
> Should we close this PR for now? If it doesn't do anything maybe we can leave it alone

Yeah I think let's just close it for now I'm not seeing any improvements.

PR #46, opened 2020-07-17
Use warm start for early stopping

inventormc:
Add early stopping for estimators that can be warm started

### DIFF
--- a/examples/warm_start.py
+++ b/examples/warm_start.py
@@ -1,5 +1,5 @@
 """
-An example training an SGDClassifier, performing grid search
+An example training a LogisticRegression model, performing grid search
 using TuneGridSearchCV.
 
 This example uses early stopping to further improve runtimes
--- a/examples/warm_start.py
+++ b/examples/warm_start.py
@@ -4,7 +4,12 @@ using TuneGridSearchCV.
 
 This example uses early stopping to further improve runtimes
 by eliminating worse hyperparameter choices early based off
-of its average test score from cross validation.
+of its average test score from cross validation. Usually
+this will require the estimator to have `partial_fit`, but
+we use sklearn's `warm_start` parameter to do this here.
+We fit the estimator for one epoch, then `warm_start`
+to pick up from where we left off, continuing until the
+trial is early stopped or `max_iters` is reached.
 """
 
 from tune_sklearn import TuneGridSearchCV
@@ -22,7 +27,7 @@ parameter_grid = {"C": [1e-4, 1e-1, 1, 2]}
 tune_search = TuneGridSearchCV(
     clf,
     parameter_grid,
-    early_stopping="MedianStoppingRule",
+    early_stopping=True,
     max_iters=10,
 )
 tune_search.fit(x_train, y_train)
--- a/examples/warm_start.py
+++ b/examples/warm_start.py
@@ -1,0 +1,37 @@
+"""
+An example training an SGDClassifier, performing grid search
+using TuneGridSearchCV.
+
+This example uses early stopping to further improve runtimes
+by eliminating worse hyperparameter choices early based off
+of its average test score from cross validation.
+"""
+
+from tune_sklearn import TuneGridSearchCV
+from sklearn.linear_model import PassiveAggressiveClassifier
+from sklearn import datasets
+from sklearn.model_selection import train_test_split
+from ray.tune.schedulers import MedianStoppingRule
+import numpy as np
+
+digits = datasets.load_digits()
+x = digits.data
+y = digits.target
+x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)
+
+clf = PassiveAggressiveClassifier()
+parameter_grid = {"C": [1e-4, 1e-1, 1, 2]}
+
+scheduler = MedianStoppingRule(grace_period=10.0)
+
+tune_search = TuneGridSearchCV(
+    clf,
+    parameter_grid,
+    early_stopping=scheduler,
+    max_iters=10,
+)
+tune_search.fit(x_train, y_train)
+
+pred = tune_search.predict(x_test)
+accuracy = np.count_nonzero(np.array(pred) == np.array(y_test)) / len(pred)
+print(accuracy)
--- a/examples/warm_start.py
+++ b/examples/warm_start.py
@@ -11,7 +11,6 @@ from tune_sklearn import TuneGridSearchCV
 from sklearn.model_selection import train_test_split
 from sklearn.datasets import load_iris
 from sklearn.linear_model import LogisticRegression
-from ray.tune.schedulers import MedianStoppingRule
 import numpy as np
 
 x, y = load_iris(return_X_y=True)
@@ -20,12 +19,10 @@ x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)
 clf = LogisticRegression()
 parameter_grid = {"C": [1e-4, 1e-1, 1, 2]}
 
-scheduler = MedianStoppingRule(grace_period=10.0)
-
 tune_search = TuneGridSearchCV(
     clf,
     parameter_grid,
-    early_stopping=scheduler,
+    early_stopping="MedianStoppingRule",
     max_iters=10,
 )
 tune_search.fit(x_train, y_train)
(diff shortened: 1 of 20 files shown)


inventormc:
> I'm a little confused at how this warm start can help early stop, and if so how the code introduces the changes necessary to make it happen. Why does setting a few config variables and a call to `fit` within the `early_stopping=True` help speed things up (as in, how does it early stop?)?

There's a set of estimators we can early stop with right now (those that have partial_fit), but this pr attempts to add more estimators that can be fit one epoch at a time. If we can fit an estimator one epoch at a time, then we can take advantage of early stopping algorithms. This means that an estimator with warm fit might work; if we set `warm_start=True`, then fit won't fit the entire estimator at once; it'll do it one epoch at at time, giving us a chance to early stop. WIthout taking advantage of this, we'd just have to use the non early stop case and `cross_validate`, doing the whole fit procedure without being able to take advantage of early stopping algorithms.

tune_sklearn/tune_basesearch.py
```diff
@@ -404,8 +404,14 @@ def _can_early_stop(self):
             bool: if the estimator can early stop
 
         """
-        return (hasattr(self.estimator, "partial_fit")
-                and callable(getattr(self.estimator, "partial_fit", None)))
+        sklearn_module = ".".join(
+            getattr(self.estimator, "__module__", None).split(".")[:-1])
+        return (hasattr(self.estimator, "partial_fit") and callable(
+            getattr(self.estimator, "partial_fit",
+                    None))) or (hasattr(self.estimator, "warm_start")
```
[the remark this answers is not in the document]
inventormc: The partial fit stuff is what we had before. This just adds an or condition to see if there's warm_start and max_iter. It also has to ensure that it isn't an ensemble or tree, because warm_start on these types of estimators doesn't allow you to continue fitting on the same estimator (you have to add more trees or more components in order to continue fitting, which isn't what we want to do).

examples/warm_start.py
```diff
@@ -0,0 +1,35 @@
+"""
+An example training an SGDClassifier, performing grid search
```
anthonyhsyu, line 2:
```suggestion
An example training an LogisticRegression, performing grid search
```

inventormc: Haven't tried this without early stopping, but I'd expect it to, since it works the same way `partial_fit` helps us have better run times. I'll try it out

inventormc: Ran this example with and without early stopping; with early stopping is about 3-5 seconds faster each time (9s vs 12s, 9s vs 14s, 12s vs 15s). However, I think this needs a bit more research because I'm not sure if this warm start method will affect accuracy. I've seen it do worse on some estimators than others.

anthonyhsyu: Iris is a pretty small data set (150 data points) so we could see larger fluctuations in accuracy. It's encouraging to see that there is runtime improvement though! Perhaps we should increase the size of the data set to fit on

inventormc: Using `make_classification(n_samples=11000)`:
Early stop with warm_start:
Time: 19.42437195777893
Acc: 0.8554545454545455

No early stop:
Time: 31.468846797943115
Acc: 0.8931818181818182

anthonyhsyu, line 2:
```suggestion
An example training a LogisticRegression model, performing grid search
```

examples/warm_start.py
```diff
@@ -0,0 +1,35 @@
...
+from ray.tune.schedulers import MedianStoppingRule
+import numpy as np
+
+x, y = load_iris(return_X_y=True)
+x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)
+
+clf = LogisticRegression()
+parameter_grid = {"C": [1e-4, 1e-1, 1, 2]}
+
+scheduler = MedianStoppingRule(grace_period=10.0)
```
richardliaw, line 23: doesn't grace_period=10 mean it won't stop anything before 10 iterations?

tune_sklearn/_trainable.py
```diff
@@ -51,6 +53,9 @@ def _setup(self, config):
             n_splits = self.cv.get_n_splits(self.X, self.y)
             self.fold_scores = np.zeros(n_splits)
             self.fold_train_scores = np.zeros(n_splits)
+            if not hasattr(self.estimator, "partial_fit"):
+                self.estimator_config["warm_start"] = True
+                self.estimator_config["max_iter"] = 1
```
richardliaw, line 58: the user can just set this themselves right?

tune_sklearn/_trainable.py
```diff
@@ -86,8 +91,13 @@ def _train(self):
                     self.y,
                     test,
                     train_indices=train)
-                self.estimator[i].partial_fit(X_train, y_train,
-                                              np.unique(self.y))
+                if hasattr(self.estimator, "partial_fit"):
+                    self.estimator[i].partial_fit(X_train, y_train,
+                                                  np.unique(self.y))
+                else:
+                    simplefilter("ignore", category=ConvergenceWarning)
```
richardliaw, line 98: this seems a little scary; maybe output it once but catch the rest of them?

examples/warm_start.py
inventormc: Yeah we can change this

tune_sklearn/_trainable.py
inventormc: Actually I wasn't able to get this working yet I think all the errors still show.

tune_sklearn/_trainable.py
inventormc: Well these are parameters not visible to the user. I thought it'd be better for the user to not worry about whether they're using `partial_fit` or `warm_start`, just if they want to early stop. `max_iter` here is different than the `max_iters` the user sets. `max_iter` is to make sklearn only fit for one epoch, while `max_iters` (which the user can set) is the usual max number of calls to _trainable.

examples/warm_start.py
```diff
@@ -0,0 +1,32 @@
...
+x, y = load_iris(return_X_y=True)
+x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)
+
+clf = LogisticRegression()
+parameter_grid = {"C": [1e-4, 1e-1, 1, 2]}
+
+tune_search = TuneGridSearchCV(
+    clf,
+    parameter_grid,
+    early_stopping="MedianStoppingRule",
```
richardliaw, line 25: can you make this early_stopping=True?

Median..rule is not a concept that most folks know about.

examples/warm_start.py
```diff
@@ -0,0 +1,32 @@
+"""
+An example training a LogisticRegression model, performing grid search
+using TuneGridSearchCV.
+
+This example uses early stopping to further improve runtimes
+by eliminating worse hyperparameter choices early based off
+of its average test score from cross validation.
```
richardliaw, line 7: Can you describe to the user how is early stopping achieved?

tune_sklearn/_trainable.py
richardliaw: OK got it. Can you leave an implementation note as to why you have to do this?

tune_sklearn/tune_basesearch.py
```diff
@@ -433,8 +433,17 @@ def _can_early_stop(self):
             bool: if the estimator can early stop
 
         """
-        return (hasattr(self.estimator, "partial_fit")
-                and callable(getattr(self.estimator, "partial_fit", None)))
+        sklearn_module = ".".join(
+            getattr(self.estimator, "__module__", None).split(".")[:-1])
```
richardliaw, line 437: This doesn't seem like it'll work for all cases. If it returns "none", you'll call None.split, which will fail.

tune_sklearn/tune_basesearch.py
```diff
@@ -433,8 +433,17 @@ def _can_early_stop(self):
...
+
+        can_partial_fit = hasattr(self.estimator, "partial_fit") and callable(
+            getattr(self.estimator, "partial_fit", None))
+        can_warm_start = (hasattr(self.estimator, "warm_start")
+                          and hasattr(self.estimator, "max_iter")
+                          and sklearn_module != "sklearn.tree"
+                          and sklearn_module != "sklearn.ensemble")
```
richardliaw, line 444: Can you add some tests for this?

PR #47, opened 2020-07-17
add tests for all schedulers

inventormc:


anthonyhsyu:
Let's merge with master again and see how it performs

PR #49, opened 2020-07-19
Better errors and warnings

inventormc:


inventormc:
> One of the tests asserts `TuneError`, change that to `ValueError` as what's raised by scikit-learn and this should be good to go

Yup. Except I had to just look at the logger content so we're not actually raising a ValueError right now.

Issue #53, opened 2020-07-22
For evaluating multiple scores, use sklearn.model_selection.cross_validate instead


sumanthratna:
# Reproducible Example

```python
from tune_sklearn import TuneSearchCV
from sklearn.ensemble import RandomForestClassifier
from sklearn import datasets
from sklearn.model_selection import train_test_split
from scipy.stats import randint
import numpy as np

digits = datasets.load_digits()
x = digits.data
y = digits.target
x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)

clf = RandomForestClassifier(random_state=317, verbose=100)
param_distributions = {
    "n_estimators": (1, 120),
}

tune_search = TuneSearchCV(
    clf,
    param_distributions,
    scoring=(
        'homogeneity_score',
        'completeness_score',
    ),
    verbose=2,
    search_optimization='bayesian'
)

tune_search.fit(x_train, x_test)

pred = tune_search.predict(y_train)
accuracy = np.count_nonzero(
    np.array(pred) == np.array(y_test)) / len(pred)
print(accuracy)
```
results in:
```
/private/tmp/venv/lib/python3.8/site-packages/tune_sklearn/tune_basesearch.py:245: UserWarning: Early stopping is not enabled. To enable early stopping, pass in a supported scheduler from Tune and ensure the estimator has `partial_fit`.
  warnings.warn("Early stopping is not enabled. "
Redis failed to start, retrying now.
Traceback (most recent call last):
  File "pls.py", line 38, in <module>
    tune_search.fit(x_train, x_test)
  File "/private/tmp/venv/lib/python3.8/site-packages/tune_sklearn/tune_basesearch.py", line 368, in fit
    result = self._fit(X, y, groups, **fit_params)
  File "/private/tmp/venv/lib/python3.8/site-packages/tune_sklearn/tune_basesearch.py", line 290, in _fit
    self.scoring = check_scoring(self.estimator, scoring=self.scoring)
  File "/private/tmp/venv/lib/python3.8/site-packages/sklearn/utils/validation.py", line 73, in inner_f
    return f(**kwargs)
  File "/private/tmp/venv/lib/python3.8/site-packages/sklearn/metrics/_scorer.py", line 430, in check_scoring
    raise ValueError("For evaluating multiple scores, use "
[...]

PR #57, opened 2020-07-24
[docs] Remove multiple metrics reference in scoring argument for TuneSearchCV

sumanthratna:
This change only affects the docstring for the `scoring` keyword argument in `TuneSearchCV`'s initializer. 

See #53 (note: this PR makes no attempt to implement multiple metrics)

CC @inventormc and @richardliaw

### DIFF
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -80,15 +80,10 @@ class TuneSearchCV(TuneBaseSearchCV):
         n_iter (int): Number of parameter settings that are sampled.
             n_iter trades off runtime vs quality of the solution.
             Defaults to 10.
-        scoring (str, `callable`, `list`, `tuple`, `dict`
-            or None): A single string (see Scikit-Learn documentation
-            on `scoring_parameter`) or a callable
+        scoring (str, `callable`, or None): A single string (see 
+            Scikit-Learn documentation on `scoring_parameter`) or a callable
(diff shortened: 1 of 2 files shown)


tune_sklearn/tune_search.py
```diff
@@ -80,15 +80,10 @@ class TuneSearchCV(TuneBaseSearchCV):
...
-            or None): A single string (see Scikit-Learn documentation
-            on `scoring_parameter`) or a callable
+        scoring (str, `callable`, or None): A single string (see 
+            Scikit-Learn documentation on `scoring_parameter`) or a callable
             to evaluate the predictions on the test set.
-            For evaluating multiple metrics, either give a list of (unique)
-            strings or a dict with names as keys and callables as values.
-            NOTE that when using custom scorers, each scorer should return a
-            single value. Metric functions returning a list/array of values
-            can be wrapped into multiple scorers that return one value each.
```
sumanthratna, line 91: I removed, "Metric functions returning a list/array of values can be wrapped into multiple scorers that return one value each." but I'm not entirely sure if this actually hasn't been implemented in tune-sklearn.

inventormc: Yes, we haven't implemented this yet. Thanks!

anthonyhsyu:
Thank you for the catch and PR @sumanthratna!

PR #61, opened 2020-07-25
Fix memory leaks in travis

inventormc:


tune_sklearn/_trainable.py
```diff
@@ -120,7 +120,7 @@ def _train(self):
                     self.X,
                     self.y,
                     cv=self.cv,
-                    n_jobs=-1,
+                    n_jobs=1,
```
richardliaw, line 123: @inventormc the main change is here

PR #62, opened 2020-07-28
Multimetric scoring

### LINKED ISSUE #53: For evaluating multiple scores, use sklearn.model_selection.cross_validate instead

inventormc:
Resolves #53


tests/test_gridsearch.py
```diff
@@ -100,7 +100,7 @@ def test_grid_search(self):
 
         # Test exception handling on scoring
         grid_search.scoring = "sklearn"
-        with self.assertRaises(ValueError):
+        with self.assertRaises((KeyError, ValueError)):
```
anthonyhsyu, line 103: Where does the `KeyError` come from?

tests/test_gridsearch.py
```diff
@@ -130,9 +130,8 @@ def test_grid_search_no_score(self):
             grid_search.score(X, y), grid_search_no_score.score(X, y))
 
         # giving no scoring function raises an error
-        grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
         with self.assertRaises(TypeError) as exc:
-            grid_search_no_score.fit([[1]])
+            grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
```
anthonyhsyu, line 134: Why does the error raise earlier when instantiating `TuneGridSearchCV`?

tune_sklearn/tune_basesearch.py
```diff
@@ -296,7 +315,7 @@ def _fit(self, X, y=None, groups=None, **fit_params):
         classifier = is_classifier(self.estimator)
         cv = check_cv(cv=self.cv, y=y, classifier=classifier)
         self.n_splits = cv.get_n_splits(X, y, groups)
-        self.scoring = check_scoring(self.estimator, scoring=self.scoring)
+        _check_multimetric_scoring(self.estimator, self.scoring)
```
anthonyhsyu, line 318: There's a `_check_multimetric_scoring` in `__init__` -- do we need it here too?

tune_sklearn/_trainable.py
```diff
@@ -143,20 +153,22 @@ def _train(self):
...
-            self.test_accuracy = sum(scores["test_score"]) / len(
-                scores["test_score"])
-            ret["average_test_score"] = self.test_accuracy
+            for name in self.scoring:
+                for i, score in enumerate(scores["test_%s" % name]):
+                    key_str = f"split{i}_test_%s" % name
+                    ret[key_str] = score
+                self.test_accuracy = sum(scores["test_%s" % name]) / len(
+                    scores["test_%s" % name])
+                ret["average_test_%s" % name] = self.test_accuracy
```
anthonyhsyu, line 162: Hmm.. did we actually set `average_test_score` twice? This might've been something that slipped in our early implementation too

tune_sklearn/tune_search.py
```diff
@@ -85,6 +85,8 @@ class TuneSearchCV(TuneBaseSearchCV):
             evaluate the predictions on the test set.
             See https://scikit-learn.org/stable/modules/model_evaluation.html
             #scoring-parameter for all options.
+            For evaluating multiple metrics, either give a list of (unique)
```
anthonyhsyu, line 88: list or tuple?

tune_sklearn/tune_gridsearch.py
```diff
@@ -51,6 +51,8 @@ class TuneGridSearchCV(TuneBaseSearchCV):
             string or a callable to evaluate the predictions on the test set.
             See https://scikit-learn.org/stable/modules/model_evaluation.html
             #scoring-parameter for all options.
+            For evaluating multiple metrics, either give a list of (unique)
```
anthonyhsyu, line 54: list or tuple?

tune_sklearn/tune_basesearch.py
```diff
@@ -411,7 +436,12 @@ def score(self, X, y=None):
             float: computed score
 
         """
-        return self.scoring(self.best_estimator_, X, y)
+        self._check_is_fitted("score")
+        if self.is_multi:
+            scorer_name = self.refit
+        else:
+            scorer_name = "score"
+        return self.scoring[scorer_name](self.best_estimator_, X, y)
```
anthonyhsyu, line 444: Scoring always becomes a dictionary, regardless of if it's single or multi metric?

tune_sklearn/tune_gridsearch.py
inventormc: I believe they all work but yeah need to fix up this documentation. It's from sklearn though

tune_sklearn/tune_basesearch.py
inventormc: It's just because there's this test where they set the scoring after creating the object, and it won't error when they call fit with the wrong scoring. I'm not sure if we want to account for this case (where user sets scoring after making the object)

tune_sklearn/tune_basesearch.py
inventormc: Yeah because _check_multimetric_scoring always returns a dict

tests/test_gridsearch.py
inventormc: Should be because of the _check_multimetric_scoring. I think this might've always happened though.

tests/test_gridsearch.py
inventormc: Because we have `_check_multimetric_scoring` in the constructor now instead of down in fit. I think this is necessary though; makes more sense because we need to know if it's multimetric ahead of time to ensure `refit` is set properly.

tune_sklearn/_trainable.py
inventormc: Where is it set again? Am I missing something here

tests/test_gridsearch.py
anthonyhsyu: I would be wary of just changing the tests because I feel like the KeyError would have been mentioned before since sklearn supports multimetric scoring too

tune_sklearn/tune_basesearch.py
anthonyhsyu: I think we should make it consistent with scikit-learn, so whatever that entails I think should be the behavior we should try to get. I feel like we shouldn't need to change the tests to support this unless we want to intentionally change the design from scikit-learn

tune_sklearn/tune_gridsearch.py
anthonyhsyu: Ah ok nice

tune_sklearn/_trainable.py
anthonyhsyu: Never mind, I expanded out the code and they're for the two different cases (with or without early stopping).. oops

tests/test_gridsearch.py
inventormc: We could just change it back to ValueError and I think it would be fine

tune_sklearn/tune_basesearch.py
inventormc: Yeah it was intentional because like, we already need to check if the scoring is multimeric in the constructor so might as well call this method in the constructor too. I think this makes more sense but we could also just not call it until the user calls fit. @richardliaw what do you think?

richardliaw: In tune-sklearn, can users set the scoring after creating the object?

inventormc: Yeah and you can also set it in sklearn too.

richardliaw: This validates that behavior and should be fine to leave as is right?

inventormc: I think @anthonyhsyu is saying that we shouldn't error in the constructor if the scoring is not valid right?

tests/test_gridsearch.py
anthonyhsyu: Ok, let's try that and see how it goes

tune_sklearn/tune_basesearch.py
anthonyhsyu: Ok, so I checked against scikit-learn and they perform all these checks in [`fit`](https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/model_selection/_search.py#L657), probably to handle both cases (if `scoring` is set when instantiating or set after that) with one call to `_check_multimetric_scoring`. Our `refit` metric is still set in `fit`, so would it be necessary in `__init__` then?

inventormc: Yeah we could just do the refit checks there too. I'll change it to that for now then

inventormc: Actually one of the tests fail now (`test_grid_search_groups`). In one of the loops, there's two consecutive calls to `fit()`, so we do `_check_multimetric_scoring` two times. The second time, because `self.scoring` has become a dict from the first call, it'll think that it's multimeric even if it actually isn't. I think it's cleaner to do checks in __init__, but this would mean changing the behavior.

anthonyhsyu: Maybe we could use this instead? 
https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/model_selection/_search.py#L772
Basically pull it out of the dict if it's not multimetric

tune_sklearn/tune_search.py
```diff
@@ -146,10 +146,13 @@ class TuneSearchCV(TuneBaseSearchCV):
         n_trials (int): Number of parameter settings that are sampled.
             n_trials trades off runtime vs quality of the solution.
             Defaults to 10.
-        scoring (str, `callable`, or None): A single string or a callable to
-            evaluate the predictions on the test set.
+        scoring (str, list/tuple, dict, or None): A single
+            string or a callable to evaluate the predictions on the test set.
```
richardliaw, line 150: callable stil holds right?

tune_sklearn/_trainable.py
```diff
@@ -5,6 +5,7 @@
 from ray.tune import Trainable
 from sklearn.base import clone
 from sklearn.model_selection import cross_validate
+from sklearn.model_selection._validation import _aggregate_score_dicts
```
richardliaw, line 8: Let's not depend on an internal implementation like this - https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/model_selection/_validation.py#L1606 

Can you move this out into our repo?

tune_sklearn/tune_basesearch.py
```diff
@@ -15,7 +15,7 @@
 from sklearn.base import BaseEstimator
 from sklearn.utils.validation import check_is_fitted
 from sklearn.model_selection import check_cv
-from sklearn.metrics import check_scoring
+from sklearn.metrics._scorer import _check_multimetric_scoring
```
richardliaw, line 18: same with this - this is a private method and we should be careful depending on it. Let's move it out?

tune_sklearn/tune_gridsearch.py
```diff
@@ -70,7 +73,7 @@ class TuneGridSearchCV(TuneBaseSearchCV):
             For integer/None inputs, if the estimator is a classifier and ``y``
             is either binary or multiclass, :class:`StratifiedKFold` is used.
             In all other cases, :class:`KFold` is used. Defaults to None.
-        refit (bool, str, or `callable`):
+        refit (bool or str):
```
richardliaw, line 76: Callable is still correct here right?

tests/test_trainable.py
```diff
@@ -5,7 +5,7 @@
 
 from sklearn.datasets import make_classification
 from sklearn.linear_model import LogisticRegression, SGDClassifier
-from sklearn.metrics import check_scoring
+from sklearn.metrics._scorer import _check_multimetric_scoring
```
richardliaw, line 8: same with this - let's not depend on a private method

tests/test_randomizedsearch.py
```diff
@@ -136,6 +136,37 @@ def test_local_mode(self):
...
+            places=10)
+
+        p = tune_search.cv_results_["params"]
+        scores = tune_search.cv_results_["mean_test_accuracy"]
+        cv_best_params = [
+            params for _, params in sorted(
+                zip(scores, p),
+                key=lambda pair: pair[0],
+            )
+        ]
```
richardliaw, line 167: can we just do like, 'max'?

tune_sklearn/tune_gridsearch.py
inventormc: The only thing I changed with refit here is that it has to be the name of the scorer you want to use to evaluate which hyperparameters are the best when you do multimetric. But I was looking through the code and I didn't see any part where a callable could be used for refit. I think having callable here in the docs might just be a mistake?

richardliaw (approved):
nice!

PR #63, opened 2020-07-31
Early stopping for XGBoost + Update Readme

### LINKED ISSUE #58: [Feature request] Early stopping doesn't work with XGBoost, LightGBM or CatBoost
I'm getting the following error while setting `early_stopping=True`

    ValueError: Early stopping is not supported because the estimator does not have `partial_fit`

These could be a potential fixes:
1. XGBoost: https://github.com/dmlc/xgboost/issues/1686
2. LightGBM: https://github.com/microsoft/LightGBM/issues/2718
3. CatBoost: https://github.com/catboost/catboost/issues/464

inventormc:
This PR supports early stopping for XGBoost. We leverage the incremental learning capabilities of XGBoost:

Note that this may not necessarily improve performance but instead allows us to break the training process into multiple parts.

```python
clf = XGBClassifier(n_estimators=10, nthread=8)
base_model = None
for i in range(20):
    z = clf.fit(x_tr, y_tr, xgb_model=base_model)
    y_te = z.predict(x_te)
    print(sklearn.metrics.mean_squared_error(y_te, y_pr))
    base_model = z.get_booster()
```

resolves #58


inventormc:
microsoft/LightGBM#3057 -`init_model` doesn't exist in the latest stable release yet

tune_sklearn/_trainable.py
```diff
@@ -87,8 +94,16 @@ def _train(self):
...
+                if self.is_lgbm:
+                    self.saved_models[i] = self.estimator[i].fit(
+                        X_train, y_train, init_model=self.saved_models[i])
+                elif self.is_xgb:
+                    self.estimator[i].fit(
+                        X_train, y_train, xgb_model=self.saved_models[i])
+                    self.saved_models[i] = self.estimator[i].get_booster()
+                else:
+                    self.estimator[i].partial_fit(X_train, y_train,
+                                                  np.unique(self.y))
```
richardliaw, line 106: what if you just put this in another trainable? would it make sense there?

inventormc: Why do we need to put it into another trainable?

richardliaw: i guess it's probably going to be unsustainable to have like 5 different special cases for different libraries, but right now it's probably fine

tune_sklearn/tune_basesearch.py
```diff
@@ -232,16 +236,19 @@ def __init__(self,
...
                  local_dir="~/ray_results",
-                 max_iters=None,
+                 max_iters=1,
                  use_gpu=False):
-
+        if max_iters < 1:
+            raise ValueError("max_iters must be greater than or equal to 1.")
         self.estimator = estimator
 
-        if not self._can_early_stop() and max_iters is not None:
```
inventormc, line 240: I think there's actually more cases we need to be checking here. Here's all the cases:

- User wants to early stop, and it can be done
  - **need to make sure the user sets `max_iters`, so raise a warning if this isn't done**
- User wants to early stop, and it cannot be done
  - directly throw an error, regardless of what `max_iters` is
  - **maybe it would be nice the remove the warning that would come up in the `if not self._can_early_stop() and max_iters > 1:` since there will be duplicate errors**
- User does not want to early stop, and it can be done
  - **if `max_iters` isn't set, do nothing**
  - **if `max_iters` is set, raise a warning that it is ignored because user didn't enable early stop**
- User does not want to early stop, and it cannot be done
  - **if `max_iters` isn't set, do nothing**
  - **if `max_iters` is set, raise a warning that it is ignored because user didn't enable early stop**

Right now

tune_sklearn/_trainable.py
```diff
@@ -112,8 +118,14 @@ def _train(self):
                     test,
                     train_indices=train)
                 if self._can_partial_fit():
-                    self.estimator_list[i].partial_fit(X_train, y_train,
-                                                       np.unique(self.y))
+                    if is_xgboost_model(self.main_estimator):
```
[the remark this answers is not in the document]
richardliaw: OK nice, I wrote a couple tests here.

tune_sklearn/tune_basesearch.py
richardliaw: It is now reduced to this:

- User wants to early stop, and it cannot be done
  - directly throw an error.

- User wants to early stop, and it can be done, and max_iters is not set
  - raise a warning that it should be set.

- regardless of whether it can be done, if user does not want to early stop and if `max_iters` is set: 
  - raise a warning that it is ignored / set to 1.

PR #81, opened 2020-08-29
Fix BOHB, change n_iter -> n_trials, fix up early stopping

### LINKED ISSUE #80: Rename n_iter => n_trials
It is confusing that n_iter and max_iter refer to two different things.

cc @inventormc

richardliaw:
You can see the diff here: https://github.com/inventormc/tune-sklearn-1/compare/warm-start...richardliaw:bobh

Here are the main changes incorporated in the PR:

1. We fix the BOHB example to actually leverage bohb properly.
2. early stopping was not calling partial_fit, so that is also fixed.
3. we rename n_iter to n_trials because n_iter is incredibly confusing (closes #80)


examples/discrete_hpbandster.py
```diff
@@ -1,29 +0,0 @@
...
-    "min_samples_leaf": CS.UniformIntegerHyperparameter(
-        "min_samples_leaf", 1, 5)
-}
-
-tune_search = TuneSearchCV(
-    RandomForestClassifier(),
-    space,
-    search_optimization="bohb",
-    n_iter=3,
-    max_iters=10)
```
richardliaw, line 25: @Yard1 I ended up removing this example because:
1. BOHB requires a form of iterative updating but
2. random forests don't have an iterative update ability (from what I can read - please let me know if this is wrong!)

Happy to add this back, but would want to understand how to make this work practically.

Yard1: I'm fine with it being removed, thanks for working on this.

Yard1:
@richardliaw So just to confirm, BOHB *requires* partial fit or warm start to work correctly, yes?

richardliaw:
Yeah, same with other early stopping mechanisms.

tune_sklearn/_trainable.py
```diff
@@ -20,6 +20,14 @@ class _Trainable(Trainable):
     and restore routines.
 
     """
+    estimator_list = None
+
+    def _can_partial_fit(self):
+        return hasattr(self.main_estimator, "partial_fit")
+
+    @property
+    def main_estimator(self):
```
anthonyhsyu, line 29: Thanks for the catch on partial_fit not found for list objects!

tune_sklearn/tune_search.py
```diff
@@ -143,8 +143,8 @@ class TuneSearchCV(TuneBaseSearchCV):
             this parameter is ignored for ``"bohb"``, as it requires
             ``HyperBandForBOHB``.
 
-        n_iter (int): Number of parameter settings that are sampled.
-            n_iter trades off runtime vs quality of the solution.
+        n_trials (int): Number of parameter settings that are sampled.
```
anthonyhsyu, line 146: This would clear confusion with `max_iters`, but do you think it'd be helpful to specify that this is the equivalent of `n_iter` from scikit-learn? I remember we discussed naming `max_iters` before too so I'm wondering if it's a problem about this parameter's name or `max_iters`'s.

richardliaw: Hm, good point. I think the reference to sklearn doesn't make sense since sklearn doesn't support "iterative models / leveraging partial fit". Since we do, it becomes a source of confusion.

PR #90, opened 2020-09-05
Warm start for ensembles

inventormc:
Add functionality to warm start ensembles estimators in sklearn. This relies on incrementing `n_estimators` one at a time to get the effect of fitting one estimator at a time in the ensemble. This unfortunately means the user can't cross validate the `n_estimators` parameter if they choose to early stop.

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -13,8 +13,9 @@ import ray.cloudpickle as cpickle
 import warnings
 
 from tune_sklearn._detect_xgboost import is_xgboost_model
-from tune_sklearn.utils import (check_warm_start_iter, check_warm_start_ensemble,
-                                check_partial_fit, _aggregate_score_dicts)
+from tune_sklearn.utils import (check_warm_start_iter,
+                                check_warm_start_ensemble, check_partial_fit,
+                                _aggregate_score_dicts)
 
 
 class _Trainable(Trainable):
@@ -140,17 +141,12 @@ class _Trainable(Trainable):
         if self.early_stopping:
             for i, (train, test) in enumerate(self.cv.split(self.X, self.y)):
                 estimator = self.estimator_list[i]
-                X_train, y_train = _safe_split(estimator, self.X,
-                                               self.y, train)
+                X_train, y_train = _safe_split(estimator, self.X, self.y,
+                                               train)
                 X_test, y_test = _safe_split(
-                    estimator,
-                    self.X,
-                    self.y,
-                    test,
-                    train_indices=train)
+                    estimator, self.X, self.y, test, train_indices=train)
                 if self._can_partial_fit():
-                    estimator.partial_fit(X_train, y_train,
-                                                       np.unique(self.y))
+                    estimator.partial_fit(X_train, y_train, np.unique(self.y))
                 elif self._is_xgb():
                     estimator.fit(
                         X_train, y_train, xgb_model=self.saved_models[i])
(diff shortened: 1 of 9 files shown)


inventormc:
Still need to add some tests and documentation

tune_sklearn/utils.py
```diff
@@ -17,6 +17,14 @@ def check_warm_start(estimator):
             and is_not_tree_subclass)
 
 
+def check_warm_start_ensemble(estimator):
+    from sklearn.ensemble import BaseEnsemble
+    is_ensemble_subclass = issubclass(type(estimator), BaseEnsemble)
+
```
richardliaw, line 23: should you update the above? `check_warm_start` can probably return True if is ensemble now

inventormc: I chose to separate it because warm start on non-ensemble estimators means we set `max_iter=1`, but warm start on ensemble estimators means we set `n_estimators=1`. I thought it'd be easier to handle these different cases with two separate checks.

richardliaw: can you then rename the above to `can_warm_start_iter`?

tune_sklearn/_trainable.py
```diff
@@ -143,6 +152,14 @@ def _train(self):
...
                 elif self._can_warm_start():
                     self.estimator_list[i].fit(X_train, y_train)
+                elif self._can_warm_start_ensemble():
+                    # User will not be able to fine tune the n_estimators
+                    # parameter using ensemble early stopping
+                    updated_n_estimators = self.estimator_list[i].get_params(
+                    )["n_estimators"] + 1
+                    self.estimator_list[i].set_params(
+                        **{"n_estimators": updated_n_estimators})
+                    self.estimator_list[i].fit(X_train, y_train)
```
richardliaw, line 162: nit: rename `estimator = self.estimator_list[i]`

tune_sklearn/_trainable.py
```diff
@@ -92,6 +97,10 @@ def _setup(self, config):
                 self.estimator_config["warm_start"] = True
                 self.estimator_config["max_iter"] = 1
 
+            if not self._can_partial_fit() and self._can_warm_start_ensemble():
+                self.estimator_config["warm_start"] = True
+                self.estimator_config["n_estimators"] = 0
```
richardliaw, line 102: Add a comment about implementation?

tune_sklearn/tune_basesearch.py
```diff
@@ -405,6 +405,10 @@ def _fit(self, X, y=None, groups=None, **fit_params):
                 mode="max",
                 scope="last")
             self.best_params = self._clean_config_dict(best_config)
+            if not check_partial_fit(
```
richardliaw, line 408: BTW, I think we should actually log something here. I think this behavior is somewhat of a "surprise", so being upfront that this methodology is being leveraged would be nice to mention.

tune_sklearn/tune_basesearch.py
```diff
@@ -405,6 +406,14 @@ def _fit(self, X, y=None, groups=None, **fit_params):
...
+                    self.estimator) and check_warm_start_ensemble(
+                        self.estimator):
+                logger.info("Warm start uses `n_estimators` to warm "
```
richardliaw, line 412:
```suggestion
                logger.info("tune-sklearn uses `n_estimators` to warm "
```

richardliaw (approved):

Yard1:
Wouldn't it be a good idea to explicitly raise an exception if the user tries to tune `n_estimators` with early stopping? I worry that warnings will be unnoticed.

richardliaw:
@yard1 Yes, that's a good idea. Let me make a quick issue.

PR #102, opened 2020-09-13
Raise errors for n_estimators and auto set parameters

### LINKED ISSUE #101: Explicitly raise an exception if the user tries to tune `n_estimators` with early stopping
> Wouldn't it be a good idea to explicitly raise an exception if the user tries to tune `n_estimators` with early stopping? I worry that warnings will be unnoticed.

_Originally posted by @Yard1 in https://github.com/ray-project/tune-sklearn/pull/90#issuecomment-691522918_

This should perhaps be raised in `base_search`. We may also want to raise exceptions for other automatic parameter settings too.

cc @inventormc

inventormc:
Resolve #101

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -95,6 +95,10 @@ class _Trainable(Trainable):
                 # max_iter is to make sklearn only fit for one epoch,
                 # while max_iters (which the user can set) is the usual max
                 # number of calls to _trainable.
+                if "max_iter" in self.estimator_config:
+                    raise ValueError("tune-sklearn uses `max_iter` to warm "
+                                     "start, so this parameter can't be "
+                                     "set when warm start early stopping. ")
                 self.estimator_config["warm_start"] = True
                 self.estimator_config["max_iter"] = 1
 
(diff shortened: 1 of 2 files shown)


tune_sklearn/_trainable.py
```diff
@@ -95,6 +95,10 @@ def _setup(self, config):
                 # max_iter is to make sklearn only fit for one epoch,
                 # while max_iters (which the user can set) is the usual max
                 # number of calls to _trainable.
+                if "max_iter" in self.estimator_config:
+                    raise ValueError("tune-sklearn uses `max_iter` to warm "
+                                     "start, so this parameter can't be "
+                                     "set when warm start early stopping. ")
```
richardliaw, line 101: Instead of `_trainable` do you think we can do this in `tune_basesearch.__init__`?

This is so that we can raise an error faster (Before starting Ray).

inventormc: We'd have to do it in the subclasses because the parameter dict that we have to check aren't passed to the superclass. I thought it would be nicer to do it when they call fit so we don't have to have duplicate code in the subclasses.

richardliaw: ah good point - i guess my main concern is that the ray tune stacktrace is not as nice as if you just raised it in the local class (there's like a bunch of raylet stuff and pyx stuf). You could also define a helper method that encapsulates the ValueError logic right?

PR #103, opened 2020-09-13
Early stopping support for sklearn Pipelines

Yard1:
Allows to use early stopping on sklearn Pipelines. The last step (almost always the estimator) will be treated to the same early stopping checks as non-Pipeline estimators.

* Adds a new param to `TuneSearchCV` and `TuneGridSearchCV`, `pipeline_detection` (open to better name suggestions). If True, will handle early stopping if the estimator is a Pipeline.
* Added a new `_Trainable` child class, `_PipelineTrainable`, handling early stopping for Pipelines.
* Added an example.

@richardliaw

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -413,9 +413,7 @@ class _PipelineTrainable(_Trainable):
         """
         # User will not be able to fine tune the n_estimators
         # parameter using ensemble early stopping
-        updated_n_estimators = estimator.get_params(
-        )[f"{self.base_estimator_name}__n_estimators"] + 1
-        estimator.set_params(**{
-            f"{self.base_estimator_name}__n_estimators": updated_n_estimators
-        })
+        n_estimator_key = f"{self.main_estimator_name}__n_estimators"
+        updated_n_estimators = estimator.get_params()[n_estimator_key] + 1
+        estimator.set_params(**{n_estimator_key: updated_n_estimators})
         estimator.fit(X_train, y_train)
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -86,25 +86,7 @@ class _Trainable(Trainable):
         self.saved_models = []  # XGBoost specific
 
         if self.early_stopping:
-            assert self._can_early_start()
-            n_splits = self.cv.get_n_splits(self.X, self.y)
-            self.fold_scores = np.empty(n_splits, dtype=dict)
-            self.fold_train_scores = np.empty(n_splits, dtype=dict)
-            if not self._can_partial_fit() and self._can_warm_start_iter():
-                # max_iter here is different than the max_iters the user sets.
-                # max_iter is to make sklearn only fit for one epoch,
-                # while max_iters (which the user can set) is the usual max
-                # number of calls to _trainable.
-                self.estimator_config["warm_start"] = True
-                self.estimator_config["max_iter"] = 1
-
-            if not self._can_partial_fit() and self._can_warm_start_ensemble():
-                # Each additional call on a warm start ensemble only trains
-                # new estimators added to the ensemble. We start with 0
-                # and add an estimator before each call to fit in _train(),
-                # training the ensemble incrementally.
-                self.estimator_config["warm_start"] = True
-                self.estimator_config["n_estimators"] = 0
+            n_splits = self._setup_early_stopping()
 
             for i in range(n_splits):
                 self.estimator_list[i].set_params(**self.estimator_config)
@@ -113,6 +95,30 @@ class _Trainable(Trainable):
                 self.saved_models = [None for _ in range(n_splits)]
         else:
             self.main_estimator.set_params(**self.estimator_config)
+
+    def _setup_early_stopping(self):
+        assert self._can_early_start()
+        n_splits = self.cv.get_n_splits(self.X, self.y)
+        self.fold_scores = np.empty(n_splits, dtype=dict)
+        self.fold_train_scores = np.empty(n_splits, dtype=dict)
+        if not self._can_partial_fit():
+            if self._can_warm_start_iter():
+                # max_iter here is different than the max_iters the user sets.
+                # max_iter is to make sklearn only fit for one epoch,
+                # while max_iters (which the user can set) is the usual max
+                # number of calls to _trainable.
+                self.estimator_config["warm_start"] = True
+                self.estimator_config["max_iter"] = 1
+
+            elif self._can_warm_start_ensemble():
+                # Each additional call on a warm start ensemble only trains
+                # new estimators added to the ensemble. We start with 0
+                # and add an estimator before each call to fit in _train(),
+                # training the ensemble incrementally.
+                self.estimator_config["warm_start"] = True
+                self.estimator_config["n_estimators"] = 0
+
+        return n_splits
 
     def step(self):
         # forward-compatbility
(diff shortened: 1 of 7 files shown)


tune_sklearn/_trainable.py
```diff
@@ -314,3 +320,102 @@ def reset_config(self, new_config):
...
+
+    Overrides Ray Tune's Trainable class to specify the setup, train, save,
+    and restore routines.
+
+    Runs all checks and configures parameters for the last step of the Pipeline.
+
+    """
+
+    def _can_partial_fit(self):
+        return check_partial_fit(self.main_estimator.steps[-1][1])
```
richardliaw, line 336: what if you just overrode `self.main_estimator`?

richardliaw: def main_estimator(self):
    return self.main_estimator.steps[-1][1]

tune_sklearn/_trainable.py
```diff
@@ -314,3 +320,102 @@ def reset_config(self, new_config):
...
+            X_train_transformed = estimator.fit_transform(X_train, y_train)
+            estimator.steps[-1] = (estimator.steps[-1][0], last_step)
+            estimator.steps[-1][1].partial_fit(X_train_transformed, y_train,
+                                               np.unique(self.y))
+
+    def _early_stopping_xgb(self, i, estimator, X_train, y_train):
+        """Handles early stopping on XGBoost estimators.
+
+        """
+        estimator.fit(X_train, y_train, {
```
richardliaw, line 397: do you need "**" before the dict?

tune_sklearn/_trainable.py
```diff
@@ -314,3 +320,102 @@ def reset_config(self, new_config):
...
+            f"{self.main_estimator.steps[-1][0]}__xgb_model": self.
+            saved_models[i]
+        })
+        self.saved_models[i] = estimator.get_booster()
+
+    def _early_stopping_iter(self, i, estimator, X_train, y_train):
+        """Handles early stopping on estimators supporting `warm_start`.
+
+        """
+        estimator.fit(X_train, y_train)
```
richardliaw, line 407: this is the same as the parent class right? if so maybe we can just remove this?

tune_sklearn/_trainable.py
```diff
@@ -314,3 +320,102 @@ def reset_config(self, new_config):
...
+
+        """
+        # User will not be able to fine tune the n_estimators
+        # parameter using ensemble early stopping
+        updated_n_estimators = estimator.get_params(
+        )[f"{self.main_estimator.steps[-1][0]}__n_estimators"] + 1
+        estimator.set_params(
+            **{
+                f"{self.main_estimator.steps[-1][0]}__n_estimators": updated_n_estimators
+            })
```
richardliaw, line 420: for _PipelineTrainable, how about let's do:

```
@property
def main_estimator_name(self):
    return self.main_estimator.steps[-1][0]

Comment (richardliaw) [reply tune_sklearn/_trainable.py:420]: so then this becomes:

```suggestion
        n_estimator_key = f"{self.main_estimator_name}__n_estimators"
        updated_n_estimators = estimator.get_params()[n_estimator_key] + 1
        estimator.set_params(
            **{n_estimator_key: updated_n_estimators})
```

tune_sklearn/_trainable.py
Yard1: Good catch, will do

tune_sklearn/_trainable.py
Yard1: Whoops

tune_sklearn/_trainable.py
Yard1: Yep, looks much better that way, thanks

tune_sklearn/_trainable.py
Yard1: Done. Named it `base_estimator` to avoid conflicts

tune_sklearn/_trainable.py
```diff
@@ -314,3 +320,100 @@ def reset_config(self, new_config):
...
+        return check_partial_fit(self.base_estimator)
+
+    def _can_warm_start_iter(self):
+        return check_warm_start_iter(self.base_estimator)
+
+    def _can_warm_start_ensemble(self):
+        return check_warm_start_ensemble(self.base_estimator)
+
+    def _is_xgb(self):
+        return is_xgboost_model(self.base_estimator)
```
richardliaw, line 353: since the property is overriden you'll be able to remove these right?

Yard1: It has a different name now. I didn't want to override it completely, because there are places where the entire Pipeline needs to be used.

Yard1: In fact, the entire Pipeline should be used by default, and the last step should be only used in specific places

tune_sklearn/tune_basesearch.py
```diff
@@ -253,10 +254,19 @@ def __init__(self,
...
         if max_iters < 1:
             raise ValueError("max_iters must be greater than or equal to 1.")
         self.estimator = estimator
+        self.base_estimator = estimator
+        self.base_estimator_name = ""
+        self.pipeline_detection = pipeline_detection
+
+        if self.pipeline_detection and isinstance(self.base_estimator,
+                                                  Pipeline):
+            self.base_estimator_name, self.base_estimator = self.base_estimator.steps[
```
richardliaw, line 268: so pipeline_detection is only useful for the case where the user is using 1. early stopping AND 2. is using a pipeline right?

You would never want - "early stopping = True AND classifier=pipeline AND pipeline_detection=False" right? That'd be the same as:

"early stopping = False AND using pipeline"

If the above is true, then maybe we can just get rid of `pipeline_detection` and just automatically use _PipelineTrainable if "classifier=pipeline AND early stopping = True"

Let me know if that's not right!

Yard1: Basically, my intention was to allow the user to choose to opt out of this if they, for example, are using a custom Pipeline that eg. implements `partial_fit` or needs to be handled in some other way. That's why I have it set to True by default, but there's still the option not to do it.

tune_sklearn/tune_gridsearch.py
```diff
@@ -192,6 +199,8 @@ def _tune_run(self, config, resources_per_trial):
                 `tune.run`.
 
         """
+        trainable = _Trainable if not self.base_estimator_name else _PipelineTrainable
```
richardliaw, line 202: (following the above comment)

can we have a method like `check_partial_fit(self.base_estimator)` but:

`check_pipeline(self.base_estimator)`

and this line will become:

```
trainable = _Trainable 
if check_pipeline(self.base_estimator) and self. early_stopping:
    trainable = _PipelineTrainable
```
that way, `TuneSearchCV` won't need to keep track of `self.base_estimator_name`

Yard1: Added, thanks

tune_sklearn/tune_basesearch.py
richardliaw: OK got it; I guess that's fair. I was just worried about having too many parameters in the initialization. 

What do you think about renaming `pipeline_detection` to `fit_last_stage=True` (and say this is only relevant if the estimator is a Pipeline object)?

tune_sklearn/_trainable.py
richardliaw: Got it!

tune_sklearn/tune_gridsearch.py
```diff
@@ -192,6 +200,10 @@ def _tune_run(self, config, resources_per_trial):
                 `tune.run`.
 
         """
+        trainable = _Trainable
+        if check_is_pipeline(self.estimator) and self.early_stopping:
```
richardliaw, line 204: if you plan on keeping `pipeline_detection => fit_last_stage`, don't forget to reference it here

```
if check... and self.early_stopping and self.fit_last_stage:
```
^ also for `TuneSearchCV`

richardliaw:
last comment - could you add a quick unit test or two in `test_gridsearch.py` and/or `test_randomizedsearch`?

Looks good and will merge after above comments + a unit test. Thanks a bunch @Yard1 !

tune_sklearn/tune_gridsearch.py
Yard1: Hmm, `fit_last_stage` doesn't quite sound right either. Maybe just `pipeline_early_stopping`?

richardliaw: `early_stop_last_stage` ? pipeline_early_stopping=False sounds like no early stopping will happen for Pipelines, but in fact the point of enabling it is to allow early_stopping to happen for `Pipeline` objects with partial fit right?

Yard1: It's actually the other way. It's supposed to handle pipelines *without* `partial_fit` (like the default sklearn one which is used by 99% of people). `coerce_pipeline`? convert doesn't seem right as there isn't any conversion involved.

richardliaw: Got it, let me try to recap:

This is currently what we have today:
```
pipeline_detection (bool): Only relevant if estimator is Pipeline object and early_stopping is enabled/True. If True, early stopping will only be performed taking the last step of the pipeline into account. If False, early stopping will be determined by 'Pipeline.warm_start' or 'Pipeline.partial_fit' capabilities, which is not supported by standard SKlearn.
```

what I'm proposing is:
```
pipeline_auto_early_stop [or variant] (bool): Only relevant if estimator is Pipeline object and early_stopping is enabled/True. If True, early stopping will be performed on the last stage of the pipeline (which must support early stopping). If False, early stopping will be determined by 'Pipeline.warm_start' or 'Pipeline.partial_fit' capabilities, which is by default not supported by standard SKlearn.
```

`pipeline_auto_early_stop` actually works for me - let me know if that sounds good to you!

richardliaw:
looks like a couple tests are failing - let me know if you run into any trouble fixing those!

Yard1:
@richardliaw everything aside one Mac job passes.

richardliaw (approved):

richardliaw:
@Yard1 awesome!

PR #105, opened 2020-09-18
Reformat early stopping with enum

inventormc:
Group checks for early stopping type into a util function that returns an enum indicating the early stop type. Will help prevent the need for redundant checks in other places.

### DIFF
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -415,13 +415,19 @@ class TuneBaseSearchCV(BaseEstimator):
             best_config = analysis.get_best_config(
                 metric=metric, mode="max", scope="last")
             self.best_params = self._clean_config_dict(best_config)
+
+            self.best_estimator_ = clone(self.estimator)
             if self.early_stop_type == EarlyStopping.warm_start_ensemble:
                 logger.info("tune-sklearn uses `n_estimators` to warm "
                             "start, so this parameter can't be "
                             "set when warm start early stopping. "
                             "`n_estimators` defaults to `max_iters`.")
-                self.best_params["n_estimators"] = self.max_iters
-            self.best_estimator_ = clone(self.estimator)
+                if check_is_pipeline(self.estimator):
+                    cloned_base_estimator = self.best_estimator_.steps[-1][1]
+                    cloned_base_estimator.set_params(
+                        **{"n_estimators": self.max_iters})
+                else:
+                    self.best_params["n_estimators"] = self.max_iters
             self.best_estimator_.set_params(**self.best_params)
             self.best_estimator_.fit(X, y, **fit_params)
 
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -22,12 +22,15 @@ import ray
 from ray.tune.schedulers import (
     PopulationBasedTraining, AsyncHyperBandScheduler, HyperBandScheduler,
     MedianStoppingRule, TrialScheduler, ASHAScheduler)
+from ray.tune.logger import (TBXLogger, JsonLogger, CSVLogger, MLFLowLogger,
+                             Logger)
 from ray.tune.error import TuneError
 import numpy as np
 from numpy.ma import MaskedArray
 import warnings
 import multiprocessing
 import os
+import inspect
 
 from tune_sklearn.utils import (EarlyStopping, get_early_stop_type,
                                 check_is_pipeline, _check_multimetric_scoring)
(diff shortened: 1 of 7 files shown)


tune_sklearn/tune_basesearch.py
```diff
@@ -414,15 +415,19 @@ def _fit(self, X, y=None, groups=None, **fit_params):
...
+
+            self.best_estimator_ = clone(self.estimator)
+            if self.early_stop_type == EarlyStopping.warm_start_ensemble:
                 logger.info("tune-sklearn uses `n_estimators` to warm "
                             "start, so this parameter can't be "
                             "set when warm start early stopping. "
                             "`n_estimators` defaults to `max_iters`.")
-                self.best_params["n_estimators"] = self.max_iters
-            self.best_estimator_ = clone(self.estimator)
+                if check_is_pipeline(self.estimator):
```
inventormc, line 425: @Yard1 I think this was accidentally left out in the early stopping pipeline pr. Could you just confirm that this is correct?

For ensemble early stopping, because we add 1 to `n_estimators` and then call fit for each call to _trainable to incrementally train ensembles, the final `n_estimators` is the same as the number of calls to _trainable (which should be `max_iters` for the best set of hyperparameters). However, you can't set this directly for pipelines, so I'm trying to set it to the base estimator here. Is it okay to index in like this and set `n_estimators`? Or do I need to add it to the `self.best_params` dict using the underscore syntax (e.g `"classify__min_samples_split": [2, 3, 4]`) and set it for the overall pipeline.

Yard1: No, it is ok to do it like this. You can modify each pipeline step separately without issues. I opted to use the underscore syntax due to preference, but either works.

tune_sklearn/utils.py
```diff
@@ -30,6 +40,40 @@ def check_warm_start_ensemble(estimator):
...
+def get_early_stop_type(estimator, early_stopping):
+    if not early_stopping:
+        return EarlyStopping.no_early_stop
+    can_partial_fit = check_partial_fit(estimator)
+    can_warm_start_iter = check_warm_start_iter(estimator)
+    can_warm_start_ensemble = check_warm_start_ensemble(estimator)
+    is_xgb = is_xgboost_model(estimator)
+
+    if can_partial_fit:
+        return EarlyStopping.partial_fit
```
richardliaw, line 66: EarlyStopping.PARTIAL_FIT

tune_sklearn/utils.py
```diff
@@ -30,6 +40,40 @@ def check_warm_start_ensemble(estimator):
...
 
+def check_error_warm_start(estimator, estimator_config):
+    early_stop_type = get_early_stop_type(estimator)
+    if (early_stop_type == EarlyStopping.WARM_START_ITER
+            and "max_iter" in estimator_config):
+        raise ValueError("tune-sklearn uses `max_iter` to warm "
+                         "start, so this parameter can't be "
+                         "set when warm start early stopping. ")
+    if (early_stop_type == EarlyStopping.WARM_START_ENSEMBLE
+            and "n_estimators" in estimator_config):
```
Yard1, line 51: Also check for `max_iter` and `n_estimators` in pipeline config. Something like:
```python
if check_is_pipeline(estimator):
    if (early_stop_type == EarlyStopping.WARM_START_ENSEMBLE
            and f"{estimator.steps[-1][0]}__n_estimators" in estimator_config):
```

tune_sklearn/utils.py
```diff
@@ -30,6 +40,40 @@ def check_warm_start_ensemble(estimator):
...
+                         "start, so this parameter can't be "
+                         "set when warm start early stopping. ")
+
+
+def get_early_stop_type(estimator, early_stopping):
+    if not early_stopping:
+        return EarlyStopping.NO_EARLY_STOP
+    can_partial_fit = check_partial_fit(estimator)
+    can_warm_start_iter = check_warm_start_iter(estimator)
+    can_warm_start_ensemble = check_warm_start_ensemble(estimator)
```
Yard1, line 62: Need pipeline checks here. It would suffice to just call those `check_` methods on the final pipeline estimator.

Yard1 (changes requested):

tune_sklearn/utils.py
inventormc: I think this should be checking on the final pipeline estimator right? In tune_basesearch this helper method is called on `self.base_estimator`

tune_sklearn/utils.py
inventormc: Ah good catch

inventormc: Fixed in #106

tune_sklearn/utils.py
Yard1: Yes, that is correct. `estimator.steps[-1][1]` if `estimator` is a pipeline.

richardliaw:
@inventormc seems like there's a merge conflict - can you rebase?

richardliaw:
closing because seems to be redudnat with #106

PR #106, opened 2020-09-18
Warm start errors

inventormc:
Explicitly error when users try to set parameters that can't be set with early stopping (`n_estimators` for ensemble warm starting and `max_iter` for iteration warm starting).

### DIFF
--- a/tune_sklearn/utils.py
+++ b/tune_sklearn/utils.py
@@ -40,8 +40,7 @@ def check_warm_start_ensemble(estimator):
             and hasattr(estimator, "n_estimators") and is_ensemble_subclass)
 
 
-def check_error_warm_start(estimator, estimator_config):
-    early_stop_type = get_early_stop_type(estimator)
+def check_error_warm_start(early_stop_type, estimator_config):
     if (early_stop_type == EarlyStopping.warm_start_iter
             and "max_iter" in estimator_config):
         raise ValueError("tune-sklearn uses `max_iter` to warm "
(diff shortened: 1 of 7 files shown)


tune_sklearn/utils.py
```diff
@@ -1,6 +1,16 @@
 from sklearn.metrics import check_scoring
 from sklearn.pipeline import Pipeline
+from tune_sklearn._detect_xgboost import is_xgboost_model
 import numpy as np
+from enum import Enum
+
+
+class EarlyStopping(Enum):
```
richardliaw, line 8: docstring?

tune_sklearn/utils.py
```diff
@@ -1,6 +1,16 @@
...
+    PARTIAL_FIT = 1
+    WARM_START_ITER = 2
+    WARM_START_ENSEMBLE = 3
+    XGB = 4
+    NO_EARLY_STOP = 5
```
richardliaw, line 13: prefer string value instead of numeric (for better readability when serialized/debugged)

richardliaw (approved):
wow this is clean! left a couple comments for the treatment of the enum, but feel free to merge when addressed!

richardliaw:
What's the diff between this and #105 ?

inventormc:
> What's the diff between this and #105 ?

This was supposed to only handle the errors for warm start, but I had to take commits from #105 since this relies on the enum changes in that pr.

PR #109, opened 2020-09-28
Early stopping support for LightGBM and CatBoost

Yard1:
Adds early stopping support to LightGBM and CatBoost using the same principle as the existing support for XGBoost.

Also replaced hardcoded Enum values with `Enum.auto()`.

Unit tests and an example are included.


Yard1:
Only Mac failing, again.

tune_sklearn/_trainable.py
```diff
@@ -104,7 +105,10 @@ def _early_stopping_partial_fit(self, i, estimator, X_train, y_train):
         """Handles early stopping on estimators that support `partial_fit`.
 
         """
-        estimator.partial_fit(X_train, y_train, np.unique(self.y))
+        try:
+            estimator.partial_fit(X_train, y_train, classes=np.unique(self.y))
+        except AttributeError:
```
inventormc, line 110: Where does this `AttributeError` come from?

tune_sklearn/utils.py
```diff
@@ -1,16 +1,19 @@
...
-    WARM_START_ENSEMBLE = 3
-    XGB = 4
-    NO_EARLY_STOP = 5
+    PARTIAL_FIT = auto()
+    WARM_START_ITER = auto()
+    WARM_START_ENSEMBLE = auto()
+    XGB = auto()
+    LGBM = auto()
+    CATBOOST = auto()
+    NO_EARLY_STOP = auto()
```
inventormc, line 16: Nice!

inventormc (dismissed):
Mostly looks good to me, just a few small things

tune_sklearn/_trainable.py
Yard1: Some estimators don't support classes in partial fit.

examples/catboostclassifier.py
```diff
@@ -0,0 +1,33 @@
...
+    "learning_rate": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
+    "depth": [3, 4, 5],
+}
+
+catboostclf = CatBoostClassifier(n_estimators=50, )
+
+digit_search = TuneSearchCV(
+    catboostclf,
+    param_distributions=params,
+    n_trials=3,
```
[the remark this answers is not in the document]
Yard1: Whoops

tune_sklearn/_trainable.py
richardliaw: hmm @Yard1 how about using `inspect.getfullargspec` to detect if classes is in the method signature rather than eagerly executing?

tune_sklearn/utils.py
richardliaw: wow

richardliaw:
looks good to me, @Yard1 ! great work - just one comment about `inspect`. @inventormc can you merge when it looks good?

tune_sklearn/_trainable.py
Yard1: I don't think there is any benefit to what you propose. There aren't going to be any side effects from the execution as the AttributeError will be raised before the function body. Using `inspect` just seems overkill here, especially considering there are no other attributes that could raise this exception. Plus this way it's kept simple and without the need for importing anything.

richardliaw: Well the benefit is that if the user's estimator.partial_fit is implemented incorrectly, any underlying attribute error will not be accidentally caught. 

inspect is stdlib so there's no harm in importing it.

richardliaw: AFAIK inspect is the standard way of doing this - https://stackoverflow.com/questions/23228664/how-to-check-which-arguments-a-function-method-takes

Yard1: Very well then, will do 👍

Yard1:
@richardliaw @inventormc Inspect implemented!

richardliaw (approved):

Issue #110, opened 2020-09-28
Multi-metric scoring with non-default `search_optimization` values


mattKretschmer:
When using multi-metric scoring with non-default values for `search_optimization` in `TuneSearchCV`, I was getting a `KeyError` when the refitting process was happening. I've included a code snipped at a screenshot of the error below. I had already successfully installed `scikit-optimize` and `hyperopt` already.

A code-snipped that reproduces this error is:
```
from tune_sklearn import TuneSearchCV
import numpy as np
from sklearn.datasets import make_regression
from sklearn.model_selection import train_test_split
from xgboost import XGBRegressor

# Set training and validation sets
X, y = make_regression(n_samples=100, n_features=10, n_informative=5)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=20)

# Example parameters to tune from xgb
parameters = {
                "regressor__max_depth": [2,3,5,8],
                "regressor__learning_rate": (0.05, 0.2, 'uniform')
}

tune_search = TuneSearchCV(
    XGBRegressor(random_state=42, objective='reg:squarederror',
                n_estimators=1000, verbose=False, n_jobs = 1),
    parameters,
    n_jobs=1,
    refit='mae',
    search_optimization='bayesian',
    n_trials = 2,
    scoring = {
                "mae": "neg_mean_absolute_error",
                "rmse": "neg_root_mean_squared_error",
                "r2": "r2",
    }
)
tune_search.fit(X_train, y_train)
end = time.time()
```
<img width="903" alt="tune_sklearn_keyerror" src="https://user-images.githubusercontent.com/11320792/94465426-e17cc880-018d-11eb-8d37-7e691cdb6cc9.png">

richardliaw:
Closed by #111 !

PR #111, opened 2020-09-28
Fix of multi_metric scoring issue

mattKretschmer:
This is a fix for Issue #110, which looked to be caused by hard-coded metric names when calling other optimizers. Moving the multi_metric scoring check to the constructor (rather than in `_fit`) helped.

### DIFF
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -131,9 +131,8 @@ class GridSearchTest(unittest.TestCase):
             grid_search.score(X, y), grid_search_no_score.score(X, y))
 
         # giving no scoring function raises an error
-        grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
         with self.assertRaises(TypeError) as exc:
-            grid_search_no_score.fit([[1]])
+            grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
 
         self.assertTrue("no scoring" in str(exc.exception))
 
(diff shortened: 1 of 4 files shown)


tests/test_gridsearch.py
```diff
@@ -131,9 +131,8 @@ def test_grid_search_no_score(self):
             grid_search.score(X, y), grid_search_no_score.score(X, y))
 
         # giving no scoring function raises an error
-        grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
         with self.assertRaises(TypeError) as exc:
-            grid_search_no_score.fit([[1]])
+            grid_search_no_score = TuneGridSearchCV(clf_no_score, {"C": Cs})
```
inventormc, line 135: In the first pr for multimetric scoring, we wanted to adhere to the sklearn behavior so that this errors when the user calls fit. I'm aware that it's now necessary to do these checks in init because the metric name is used in the `TuneSearchCV` subclass. Do we think this different behavior is okay? @richardliaw @anthonyhsyu

inventormc (commented):
Good catches! Mostly looks good to me besides one small thing.

tests/test_gridsearch.py
richardliaw: I think it's OK, makes sense to raise earlier!

richardliaw (approved):
awesome work @mattKretschmer ! Thanks for the PR!

mkretsch327:
Thank YOU! Great package, and happy to help.

PR #114, opened 2020-10-01
[WIP] `best_params` w/o `refit`, API in line with sklearn

Yard1:
This PR makes `TuneBaseSearchCV` match sklearn's `BaseSearchCV` better by allowing to access various fit properties without the need of `refit=True`, as well as adds properties that were missing from `TuneBaseSearchCV` for full parity with `BaseSearchCV`.

Code is done, but it could use some extra unit tests for multimetric scoring etc. Will do that soon.

### DIFF
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -31,6 +31,8 @@ import warnings
 import multiprocessing
 import os
 import inspect
+import time
+import numbers
 
 from tune_sklearn.utils import (EarlyStopping, get_early_stop_type,
                                 check_is_pipeline, _check_multimetric_scoring)
@@ -124,8 +126,24 @@ class TuneBaseSearchCV(BaseEstimator):
         specified.
 
         """
-        self._check_if_refit("best_params_")
+        self._check_is_fitted("best_params_")
         return self.best_params
+
+    @property
+    def best_index_(self):
+        """int: The index (of the ``cv_results_`` arrays) 
+        which corresponds to the best candidate parameter setting.
+
+        The dict at ``search.cv_results_['params'][search.best_index_]`` gives
+        the parameter setting for the best model, that gives the highest
+        mean score (``search.best_score_``).
+
+        For multi-metric evaluation, this is present only if ``refit`` is
+        specified.
+
+        """
+        self._check_is_fitted("best_index_")
+        return self.best_index
 
     @property
     def best_score_(self):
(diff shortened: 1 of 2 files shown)


Yard1:
@richardliaw @inventormc Could you actually look this over right now? As I said, the code is complete, the only thing that's really missing is a test for multimetric scoring. I was hoping to get it done sooner but I got swamped by other commitments.

richardliaw:
Ah sorry @Yard1, I totally missed your message. Do feel free to ping/mark it ready as review even if you are missing a test.

tune_sklearn/tune_basesearch.py
```diff
@@ -479,29 +539,54 @@ def _fit(self, X, y=None, groups=None, **fit_params):
...
+            # of the params are estimators as well.
+            self.best_estimator = clone(
+                base_estimator.set_params(**self.best_params))
+            refit_start_time = time.time()
+            if y is not None:
+                self.best_estimator.fit(X, y, **fit_params)
+            else:
+                self.best_estimator.fit(X, **fit_params)
+            refit_end_time = time.time()
+            self.refit_time = refit_end_time - refit_start_time
```
richardliaw, line 589: Can you be sure to set an instance variable in the constructor for this?

tune_sklearn/tune_basesearch.py
```diff
@@ -479,29 +539,54 @@ def _fit(self, X, y=None, groups=None, **fit_params):
...
-            self.best_score = float(best_result[metric])
+            # we clone again after setting params in case some
+            # of the params are estimators as well.
+            self.best_estimator = clone(
+                base_estimator.set_params(**self.best_params))
+            refit_start_time = time.time()
+            if y is not None:
+                self.best_estimator.fit(X, y, **fit_params)
+            else:
+                self.best_estimator.fit(X, **fit_params)
```
richardliaw, line 587: great catch re: unsupervised

richardliaw (commented):
This looks great. Sorry again for reviewing so late! Let's be sure to add unit tests for:

1. refit=False/True, multi=false/true
3. refit=True for unsupervised?
3. do we have a test for callable refit already?

tune_sklearn/tune_basesearch.py
Yard1: Will do!

Yard1:
@richardliaw I think there was a test for 3, will need to check again. It's been a while. I will add the rest of the tests in the coming days, and then remove the draft status once I am done.

Yard1:
@richardliaw I have looked over the tests. It looks like they already check for everything that's needed (after the small changes I have made to them earlier on). I don't think we need to add any new tests. I think it's ready for merge now.

richardliaw:
BTW, I know check_is_fitted is supposed to validate whether refit is called before the proper variable, but I think right now, the multi-metric has some error messages that could benefit from `check_is_fitted`/validation.

For example:

```python
import numpy as np
from tune_sklearn import TuneGridSearchCV
from sklearn.linear_model import Ridge, SGDClassifier
parameter_grid = {"alpha": [1e-4, 1e-1, 1], "epsilon": [0.01, 0.1]}


X = np.array([[-1, -1], [-2, -1], [1, 1], [2, 1], [2, 1]])
y = np.array([1, 1, 2, 2, 2])

tune_search = TuneGridSearchCV(
    SGDClassifier(),
    parameter_grid,
    scoring=("accuracy", "f1_micro"),
    max_iters=20,
    cv=2,
    refit=False)
tune_search.fit(X, y)
print(tune_search.best_score_)
```
And I get:
```
/Users/rliaw/dev/tune-sklearn/tune_sklearn/tune_basesearch.py:380: UserWarning: max_iters is set > 1 but incremental/partial training is not enabled. To enable partial training, ensure the estimator has `partial_fit` or `warm_start` and set `early_stopping=True`. Automatically setting max_iters=1.
  category=UserWarning)
Traceback (most recent call last):
  File "_test.py", line 18, in <module>
    print(tune_search.best_score_)
  File "/Users/rliaw/dev/tune-sklearn/tune_sklearn/tune_basesearch.py", line 157, in best_score_
    return self.best_score
AttributeError: 'TuneGridSearchCV' object has no attribute 'best_score'
```

This is because refit is not True and I'm using multi-metric scoring. Thus, can you help provide proper validation for multi-metric scoring in this PR (as to prevent this AttributeError, which is hard to debug/hard on the user to know if it is a user or framework error)?

Thanks!

Yard1:
@richardliaw Yeah, good call. Sklearn enforces `refit!=False` when using multimetric. I'll add that here too.

tune_sklearn/tune_basesearch.py
```diff
@@ -241,25 +301,39 @@ def _check_if_refit(self, attr):
...
 
         """
         if not self.refit:
-            msg = ("This {0} instance was initialized with refit=False. {1} "
-                   "is available only after refitting on the best "
-                   "parameters.").format(type(self).__name__, method_name)
-            raise NotFittedError(msg)
-        else:
-            check_is_fitted(self)
+            if check_refit == "multimetric":
```
Yard1, line 321: Unsure if this is the best way to do it. Other ones I see are splitting this into two functions, or using an enum to define possible values for the parameter. Thoughts?

tune_sklearn/tune_basesearch.py
```diff
@@ -307,22 +307,33 @@ def _check_is_fitted(self, method_name, check_refit=True):
         Args:
             method_name (str): String of the method name called from the user.
 
-            check_refit(bool)=True: Whether to also check for `self.refit`
-            param.
+            check_refit(bool/str)=True: Whether to also check for `self.refit`
```
richardliaw, line 310: hmm i think this doesn't render properly on the documentation
```suggestion
            check_refit (bool/str): Whether to also check for `self.refit`
```

Yard1: Thanks, changed

Issue #124, opened 2020-10-29
Feature Request: Support unified search space


richardliaw:
Tune now has a unified search space on 1.0; tune-sklearn should support that space too.

Yard1:
@richardliaw Does it mean we should get rid of the automatic conversion from `dict` to algo specific search space, and instead convert `dict`s to tune's unified search space and then pass it to the algos? Leaving already converted search spaces as-is, of course.

richardliaw:
Hmm, possible to leave that as-is. The issue is so that users can pass in their search space in the Tune API format.

Yard1:
Adding just that should be a quick and easy change. Just to be certain - we don't need to do any processing of it here, do we? I took a look at Tune code and it seems like it's all handled within the algos themselves.

richardliaw:
yep! all is done in tune

Yard1:
Happy to work on this, then.

richardliaw:
Great! I'll assign this to you, and just cc @krfricke when you're done!

richardliaw:
Awesome! This should be addressed with the new release :)

PR #128, opened 2020-11-01
Support for Tune search spaces

Yard1:
Adds support to Tune's search spaces, resolving https://github.com/ray-project/tune-sklearn/issues/124

### DIFF
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -736,7 +736,7 @@ class GridSearchTest(unittest.TestCase):
         print(grid_search)
         # Without timeout we would need over 50 seconds for this to
         # finish. Allow for some initialization overhead
-        self.assertLess(taken, 18.0)
+        self.assertLess(taken, 25.0)
 
 
 if __name__ == "__main__":
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -1,4 +1,5 @@
 import ray
+import ray.tune as tune
 from tune_sklearn import TuneGridSearchCV
 from tune_sklearn import TuneSearchCV
 import numpy as np
@@ -696,6 +697,18 @@ class GridSearchTest(unittest.TestCase):
 
         self.assertTrue(hasattr(grid_search, "cv_results_"))
         self.assertTrue(wrapped_init.call_args[1]["local_mode"])
+
+    def test_tune_search_spaces(self):
+        # Test mixed search spaces
+        clf = MockClassifier()
+        grid_search = TuneGridSearchCV(
+            clf, {
+                "foo_param": tune.grid_search([1, 2, 3]),
+                "bar_param": [1, 2]
+            },
+            refit=False,
+            cv=3)
+        grid_search.fit(X, y)
 
 
 if __name__ == "__main__":
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -715,7 +715,6 @@ class GridSearchTest(unittest.TestCase):
             cv=3)
         grid_search.fit(X, y)
 
-
     def test_timeout(self):
         clf = SleepClassifier()
         # SleepClassifier sleeps for `foo_param` seconds, `cv` times.
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -1,7 +1,7 @@
 import time
 
 import ray
-import ray.tune as tune
+from ray import tune
 from tune_sklearn import TuneGridSearchCV
 from tune_sklearn import TuneSearchCV
 import numpy as np
--- a/tests/test_gridsearch.py
+++ b/tests/test_gridsearch.py
@@ -324,7 +324,8 @@ class GridSearchTest(unittest.TestCase):
         with self.assertRaises(ValueError) as exc:
             TuneGridSearchCV(clf, param_dict)
         self.assertTrue(("Parameter grid for parameter (C) needs to"
-                         " be a list or numpy array") in str(exc.exception))
+                         " be a tune.grid_search, list or numpy array"
+                         ) in str(exc.exception))
 
         param_dict = {"C": []}
         clf = SVC()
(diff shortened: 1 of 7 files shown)


Yard1:
WIP. Will work more on it. No support for grid search yet (though do we need it?). Need to see if the tests go through. @krfricke FYI

Yard1:
Also @richardliaw looks like CI is doing `pip install -U https://s3-us-west-2.amazonaws.com/ray-wheels/latest/ray-0.9.0.dev0-cp36-cp36m-manylinux1_x86_64.whl`. Shouldn't it be switched to ray-1.0.0?

richardliaw:
hmm, can we update it to 1.1.0.dev0? (see the wheel links here https://docs.ray.io/en/latest/installation.html)

Yard1:
@krfricke How are tune search spaces handled in grid search? The only thing left for this PR is determining how to handle grid searching. I think that it would be fine to leave it as-is (that is only support a dict of lists instead of any search spaces), but I am open to suggestions.

richardliaw:
@Yard1 feel free to ping Kai again, perhaps on slack, after 48 hours :)

krfricke:
In Tune, grid search is represented by a `Categorical` domain with a `Grid` sampler. You can also use the [`Domain.is_grid()`](https://github.com/ray-project/ray/blob/master/python/ray/tune/sample.py#L50) method to determine if a domain should be grid searched. 

In Tune, grid search means that all combinations of all grid search items (if there is more than one grid search item) should be generated at least once. E.g. with `"a": tune.grid_search([1, 2])` the basic variant generator will generate two trials, one with `a=1` and one with `a=2`. The `num_samples` parameter will multiply this number, i.e. if `num_samples=2`, _four_ trials will be generated here. Does this help?

Yard1:
@krfricke So how would a user pass search spaces to be grid searched?

krfricke:
So in regular Ray Tune a user would pass something like

```
config = {
    "var": tune.grid_search([1, 2, 3]),
    "rnd": tune.uniform(0, 5)
}
tune.run(
    trainable,
    config=config)
```

With `num_samples=1` this would create three trials, one for each grid serach value (1, 2 and 3). The `rnd` value is drawn randomly for each of these trials.

Yard1:
Great, thanks. This clears it up.

Yard1:
@richardliaw @krfricke Tests are passing, Mac fails due to an outdated Ray version, I think. I'll double check, but in the meantime, it's ready for review.

It's also not working with the currently released Ray 1.0.0 as it's missing the functionality @krfricke added later.

richardliaw:
1.0.1 is out now :)

Yard1:
@richardliaw Great to hear that! I need to put on notifications for the main repo, hah

tune_sklearn/tune_gridsearch.py
```diff
@@ -176,9 +178,37 @@ def __init__(self,
...
+                        "Parameter grid for parameter ({0}) needs to"
+                        " be a tune.grid_search, list or numpy array,"
+                        " but got ({1})."
+                        " Single values need to be wrapped in a list"
+                        " with one element.".format(name, type(v)))
+
+                if len(v) == 0:
+                    raise ValueError(
+                        "Parameter values for parameter ({0}) need "
+                        "to be a non-empty sequence.".format(name))
```
richardliaw, line 210: can this be a helper function (not scoped within the class?)

richardliaw:
This looks good! though:
* it seems to be failing on mac build?
* can we update the readme/docstrings?

Also, @krfricke please take a look!

tune_sklearn/tune_gridsearch.py
Yard1: Done, moved to utils

Yard1:
@richardliaw Updated docstrings and readme. Not sure what's up with Mac, I'll look into it tomorrow. Maybe it's downloading the wrong version?

Yard1:
@krfricke Mac CI fails on `test_timeout`, could you take a look?

krfricke:
Yep, I'll look into it tomorrow

krfricke:
Seems there's some more overhead for the test_timeout (18.47 seconds instead of 18). Without the time budget the test would run for 50+ seconds. Should we just raise that limit to 25 to account for environment/MacOS overhead? Do you just want to include it in this PR since it is a 1 line change?

Yard1:
@krfricke Yeah, sure. Do you want to make the change or should I?

richardliaw:
Antoni, probably faster for yourself to change it ;)

Yard1:
@richardliaw Touché!

Yard1:
@richardliaw Mac now errors instead of failing. It was doing that with other PRs - I think it's something to do with running out of memory? I think it should be good to merge though, considering Linux passes.

tests/test_gridsearch.py
```diff
@@ -702,6 +705,18 @@ def test_local_mode(self):
...
+        clf = MockClassifier()
+        grid_search = TuneGridSearchCV(
+            clf, {
+                "foo_param": tune.grid_search([1, 2, 3]),
+                "bar_param": [1, 2]
+            },
+            refit=False,
+            cv=3)
+        grid_search.fit(X, y)
+
```
krfricke, line 719: Should this test if all combinations were seen?

krfricke (approved):
Looks good, maybe check for all grid search combinations in test.

tests/test_gridsearch.py
Yard1: Good call, added

richardliaw:
hmmm it looks like Mac tests are failing but it's not an out of memory issue?

Yard1:
@richardliaw Looks like the functionality needed for this PR is not yet in release Ray - that's what Mac uses, and it's what the exception on it is related to.

richardliaw:
Got it - a couple things

1. We can introduce new features (without regressions) not compatible with released Ray, but we must make sure the user knows what is the error.
2. We should avoid introducing a failing test into CI; it should be marked as skipped if anything.

Thanks!

Yard1:
@richardliaw Got it. I think we could just wait a while on this PR, unless we are planning to release a new version of tune-sklearn soon.

PR #134, opened 2020-11-05
Add `time_budget_s` parameter

krfricke:
This passes the `time_budget_s` parameter to `tune.run()`.

This also refactors parts of the base search `_format_results()` method to deal with missing values from trials (that e.g. have not been received due to early stopping/timeout).

### DIFF
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -664,7 +664,8 @@ class TuneSearchCV(TuneBaseSearchCV):
             raise ValueError(
                 f"Invalid search optimizer: {self.search_optimization}")
 
-        if isinstance(self.n_jobs, int) and self.n_jobs > 0:
+        if isinstance(self.n_jobs, int) and self.n_jobs > 0 \
+           and not self.search_optimization == "random":
             search_algo = ConcurrencyLimiter(
                 search_algo, max_concurrent=self.n_jobs)
             run_args["search_alg"] = search_algo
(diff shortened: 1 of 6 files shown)


tune_sklearn/tune_search.py
```diff
@@ -258,6 +258,9 @@ class TuneSearchCV(TuneBaseSearchCV):
             determined by 'Pipeline.warm_start' or 'Pipeline.partial_fit'
             capabilities, which are by default not supported by standard
             SKlearn. Defaults to True.
+        time_budget_s (int|float|datetime.timedelta): Global time budget in
+            seconds after which all trials are stopped. Can also be a
+            ``datetime.timedelta`` object.
```
Yard1, line 263: We may want to expand this docstring to explain when exactly they are stopped - at the end of the trial, at the start, in the middle, etc.

tune_sklearn/tune_search.py
```diff
@@ -591,25 +596,25 @@ def _tune_run(self, config, resources_per_trial):
...
-                reuse_actors=True,
-                verbose=self.verbose,
-                stop=stop_condition,
-                num_samples=self.num_samples,
-                config=config,
-                fail_fast=True,
-                resources_per_trial=resources_per_trial,
-                local_dir=os.path.expanduser(self.local_dir),
-                loggers=self.loggers)
+        run_args = dict(
```
Yard1, line 599: Great idea, been wanting to put the args in a dict myself at some point :+1:

Yard1 (changes requested):
Looks good! Can we also get a test for random search? Just in case

krfricke:
Added!

Yard1 (approved):

PR #140, opened 2020-11-17
Add `seed` parameter to make initial configuration sampling deterministic

### LINKED ISSUE #112: [Feature Request] Reproducible results for TuneSearchCV
Is it possible to produce reproducible results for TuneSearchCV using a random seed?

krfricke:
For reproducible results, we need to be able to set a random seed. This PR introduces the respective argument to `TuneSearchCV`. 

Please note that this mostly affects initial configuration sampling. With parallel evaluations, there's always non-determinism involved that can lead to different results of the training run.

Currently not implemented on BOHB.

Closes #112

### DIFF
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -2,6 +2,8 @@
     -- Anthony Yu and Michael Chau
 """
 import logging
+import random
+
 from tune_sklearn.utils import check_is_pipeline
 from tune_sklearn.tune_basesearch import TuneBaseSearchCV
 from tune_sklearn._trainable import _Trainable
@@ -262,6 +264,9 @@ class TuneSearchCV(TuneBaseSearchCV):
             seconds after which all trials are stopped. Can also be a
             ``datetime.timedelta`` object. The stopping condition is checked
             after receiving a result, i.e. after each training iteration.
+        seed (int): Seed used to initialize random number generators. Running
+            the same experiment with the same seed will generate the same
+            initial parameter configurations.
         **search_kwargs (Any):
             Additional arguments to pass to the SearchAlgorithms (tune.suggest)
             objects.
@@ -289,6 +294,7 @@ class TuneSearchCV(TuneBaseSearchCV):
                  loggers=None,
                  pipeline_auto_early_stop=True,
                  time_budget_s=None,
+                 seed=None,
                  **search_kwargs):
         search_optimization = search_optimization.lower()
         available_optimizations = [
@@ -372,6 +378,7 @@ class TuneSearchCV(TuneBaseSearchCV):
                 raise ValueError("Random search does not support "
                                  f"extra args: {search_kwargs}")
         self.search_optimization = search_optimization
+        self.seed = seed
         self.search_kwargs = search_kwargs
 
     def _fill_config_hyperparam(self, config):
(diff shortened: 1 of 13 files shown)


Yard1:
FYI test for BOHB is failing. Aside from that, looks good. One idea for better sklearn compatibility - do you think it would be possible to pass numpy `RandomState` instances in addition to integers - https://scikit-learn.org/stable/glossary.html#term-random-state? It should be possible to just get an integer seed from those and then just carry on with an integer.

krfricke:
This is weird, it works for me locally every time I run it, even when I alter the search space. I'll look into this tomorrow.
I didn't know about `RandomState` but will make sure that it is supported. Thanks!

krfricke:
Before I get to this, should we just use the existing `random_state` parameter for the other search algorithms?

Yard1:
I think that would be the best, honestly. I forgot that it's already there! I think a user would expect just one randomness parameter to control everything. I believe that's also how other libraries do it. I don't see much utility in splitting it into `random_state` and `seed`.

tune_sklearn/tune_search.py
```diff
@@ -262,6 +263,9 @@ class TuneSearchCV(TuneBaseSearchCV):
             seconds after which all trials are stopped. Can also be a
             ``datetime.timedelta`` object. The stopping condition is checked
             after receiving a result, i.e. after each training iteration.
+        seed (int): Seed used to initialize random number generators. Running
```
Yard1, line 266: Leftover docstring

Yard1 (changes requested):
Just one leftover docstring, rest looks perfect

krfricke:
But the tests seem to fail nevertheless. Weird. Could you run them locally and see if you get the same errors? I don't have any errors on my machine, no matter how often I run them.

krfricke:
Ah, so it's still just BOHB. Maybe there's a dependency mismatch or so. I'll see what I can find.

krfricke:
So even on a fresh env with py 3.6.10 it passes for me. I tried 10 different seeds sequentially, all passed. Any ideas? 
The only difference I see right now is that I'm on mac and the VM is on linux. I could try setting up a new env on a cluster machine later or try on my linux laptop.

Yard1:
I wouldn't be surprised if it was something to do with the OS. ConfigSpace uses Cython, so I'd imagine it's not as OS agnostic as other libraries. In the worst case, we may have to just not test for it. One thing we can try, however, is setting `n_jobs=1`, so it's ran in Local Mode. Perhaps that may be it?

Yard1 (approved):

richardliaw:
Hmm, looks like the seed test is failing on mac os py3.7?

Yard1:
It's BOHB again D:

krfricke:
Ok I think I got it - search space conversions might disregard custom seed. This PR in ray should fix this by introducing a `seed` parameter for BOHB: https://github.com/ray-project/ray/pull/12160

After that one is merged, tests with latest ray should pass. Let's defer this PR until then.

krfricke:
It was merged, I'll wait for the next nightly wheels, push an update to this PR and then fingers crossed.

tune_sklearn/tune_search.py
```diff
@@ -204,10 +206,9 @@ class TuneSearchCV(TuneBaseSearchCV):
             distributions.
             If int, random_state is the seed used by the random number
             generator;
-            If RandomState instance, random_state is the random number
-            generator;
+            If RandomState instance, a seed is sampled from random_state;
             If None, the random number generator is the RandomState instance
-            used by np.random. Defaults to None.
+            used by np.random and no seed is passed. Defaults to None.
```
richardliaw, line 211:
```suggestion
            used by np.random and no seed is provided. Defaults to None.
```

PR #144, opened 2020-11-23
Update Ray latest wheels URL

krfricke:
The link to the latest nightly build changed. We're currently building against an old version.

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -244,8 +244,7 @@ class _Trainable(Trainable):
                     groups=self.groups,
                     scoring=self.scoring,
                     return_train_score=self.return_train_score,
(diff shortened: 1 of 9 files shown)


Yard1:
@krfricke Looks like something that was changed in Ray is making tune-sklearn break - the tests are failing.

tune_sklearn/_trainable.py
```diff
@@ -244,6 +244,7 @@ def _train(self):
                     groups=self.groups,
                     scoring=self.scoring,
                     return_train_score=self.return_train_score,
+                    error_score="raise"
```
richardliaw, line 247: for some reason we didn't need to do this before..

richardliaw (approved):
Looks great :) will merge with tests passing.

Issue #143, opened 2020-11-22
Fractional values in gpu (resource_per_trial)


alokgogate:
Is it possible to give fractional values along with parameter `use_gpu` while using `TuneSearchCV` ?

[Labels]: question

richardliaw:
Hmm, currently not possible to do so, but let me tag this as a feature request (and it should be easy to implement!)

Yard1:
I can work on this.

richardliaw:
That's awesome - it should be really straightforward (see n_cpus)

Yard1:
Should we add a new parameter for that, or allow users to pass integers to `use_gpu`, in which case the same logic as for `n_jobs` will be used (see line 531 in `tune_basesearch.py`)?

richardliaw:
Hmm I think you just want to make n_jobs also configure fractional gpu (no
new param)

richardliaw:
Hey @alokgogate, @Yard1 just merged this -- could you give it a try?

```
TuneSearchCV(use_gpu=True, n_jobs=8)
```

will use 8 jobs with the proper CPU/GPU fractions :)

richardliaw:
I'll close this for now, but definitely let me know if this works for you (and feel free to reopen/open a new issue!)

alokgogate:
Hi @richardliaw, 

I tried to install it using the master branch 
the setup.py get's stuck and terminates after some time with the following error 👎 
```
Couldn't find index page for 'parameterized' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.python.org/simple/
No local packages or download links found for parameterized
error: Could not find suitable distribution for Requirement.parse('parameterized')
```

I check for package `parameterized` and have it already installed

richardliaw:
Hmm maybe theres a pip outage - can you try commenting it out on setup.py?

alokgogate:
didn't find pip outage or outage in the file

richardliaw:
sorry, can you just try commenting out "parametrize" in the setup.py?

alokgogate:
Don't see that one either

richardliaw:
hmm @alokgogate I just merged a fix, can you try: `pip install git+https://github.com/ray-project/tune-sklearn`

alokgogate:
Hi @richardliaw, this worked and was able to install and run a few experiments. I have a few observations though:

[...]

PR #145, opened 2020-11-23
Use fractional GPUs for a Ray cluster

Yard1:
Solves https://github.com/ray-project/tune-sklearn/issues/143.

I am not *entirely* sure if this is the full solution to the issue, but I assumed that it would be only applicable to a Ray cluster. Therefore, the new behavior is to query Ray for the number of available GPUs and assign a fraction of them to each trial in the exact same way CPUs are assigned.


richardliaw (approved):
nice! Looks like a safe change to me. Thanks for the fast turnaround!

PR #147, opened 2020-11-24
Move lint to GHA

richardliaw:
Removes lint from Travis and moves to GHA.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,10 +77,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.7
+    - name: Check out Ray
+      uses: actions/checkout@v2
+      with:
+        repository: ray-project/ray
+      path: ray
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        git clone https://github.com/ray-project/ray
         python -m pip install -U -r ray/doc/requirements-rtd.txt
         python -m pip install -r ray/doc/requirements-doc.txt
     - name: Install package
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,6 +97,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,9 +103,9 @@ jobs:
         python-version: 3.7
     - name: Install dependencies
       run: |
+        export LINT=1 PYTHONWARNINGS=ignore
         python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -65,4 +65,50 @@ jobs:
...
+        cd ray/doc/ && sphinx-build -q -E -W -T -b html source _build/html
+
+  test_lint:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
Yard1, line 101: Do we need to fetch the entire repo for this? The latest commit should be good. I'd just remove it and have it use all default arguments, line in `build_docs`.

.github/workflows/test.yml
```diff
@@ -65,4 +65,50 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        git clone https://github.com/ray-project/ray
```
Yard1, line 83: You can use `actions/checkout@v2` for it as well:
```
- uses: actions/checkout@v2
  with:
    repository: 'ray-project/ray'
	path: 'ray'
```
But tbh it's not a big deal, considering that it's a public repo. Maybe consider doing `--depth 1` to just fetch the latest commit instead of the entire history - it'll speed things up as we don't need anything but the latest commit to do an install (`action/checkout` does that by default).

Yard1 (changes requested):
Just a couple of nitpicks/questions

.github/workflows/test.yml
richardliaw: `format.sh` uses the merge base; so checking out 1 commit wasn't sufficient. The repo is small though.

.github/workflows/test.yml
richardliaw: Hmm, yeah this sounds like a good idea.

richardliaw:
@Yard1 do you know how to cancel existing builds of a PR if a new PR commit arrives?

.github/workflows/test.yml
Yard1: Gotcha, right.

Yard1:
@richardliaw I don't think it's possible with GHA.

richardliaw:
@Yard1 review again? should be passing now!

.github/workflows/test.yml
```diff
@@ -77,10 +77,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.7
+    - name: Check out Ray
+      uses: actions/checkout@v2
+      with:
+        repository: ray-project/ray
+      path: ray
```
Yard1: `path` needs to be indented to be on the same level as  `repository`

Yard1 (approved):

Issue #135, opened 2020-11-06
Exception Handling: seeing a `_queue.Empty` error


richardliaw:
If you run:

```
(base) ➜  tune-sklearn git:(raise-tuneerror) ✗ cat _test.py
import numpy as np
from tune_sklearn import TuneGridSearchCV
from sklearn.model_selection import GridSearchCV
from sklearn.linear_model import Ridge, SGDClassifier
parameter_grid = {"alpha": [1e-4], "epsilon": [0.01]}


X = np.array([[-1, -1], [-2, -1], [1, 1], [2, 1], [2, 1]])
y = np.array([1, 1, 2, 2, 2])


tune_search = TuneGridSearchCV(
    SGDClassifier(), parameter_grid, scoring="f1_micro")# , max_iters=20)
tune_search.fit(X, y)
print("done")
```

You get this exception:

```
(base) ➜  tune-sklearn git:(raise-tuneerror) ✗ python _test.py
File descriptor limit 256 is too low for production servers and may result in connection errors. At least 8192 is recommended. --- Fix with 'ulimit -n 8192'
Trial _Trainable_9426a_00000: Error processing event.
Traceback (most recent call last):
  File "_test.py", line 14, in <module>
    tune_search.fit(X, y)
  File "/Users/rliaw/dev/tune-sklearn/tune_sklearn/tune_basesearch.py", line 650, in fit
    result = self._fit(X, y, groups, **fit_params)
  File "/Users/rliaw/dev/tune-sklearn/tune_sklearn/tune_basesearch.py", line 551, in _fit
    analysis = self._tune_run(config, resources_per_trial)
  File "/Users/rliaw/dev/tune-sklearn/tune_sklearn/tune_gridsearch.py", line 259, in _tune_run
    loggers=self.loggers)
  File "/Users/rliaw/miniconda3/lib/python3.7/site-packages/ray/tune/tune.py", line 416, in run
    runner.step()
  File "/Users/rliaw/miniconda3/lib/python3.7/site-packages/ray/tune/trial_runner.py", line 389, in step
    self._process_events()  # blocking
  File "/Users/rliaw/miniconda3/lib/python3.7/site-packages/ray/tune/trial_runner.py", line 532, in _process_events
    self._process_trial(trial)
  File "/Users/rliaw/miniconda3/lib/python3.7/site-packages/ray/tune/trial_runner.py", line 547, in _process_trial
    result = self.trial_executor.fetch_result(trial)
[...]

PR #150, opened 2020-11-25
Avoid Loky Pickle Error

richardliaw:
It seems like in newer versions of sklearn, we are unable to use joblib inside Ray.

This change makes it so that users will avoid errors in the future.

Closes https://github.com/ray-project/tune-sklearn/issues/135


Yard1:
I believe it would be a good idea to add a warning if `sk_n_jobs` has been set by the user (in old code), to inform them that it is ignored due to issues with sklearn and defaulted to 0.

richardliaw:
ah, great point!

richardliaw:
@Yard1 raised a deprecation message instead!

Yard1 (approved):

PR #160, opened 2020-12-14
Add tests for `points_to_evaluate`

### LINKED ISSUE #73: [Feature Request] Add default model parameters to tuning trials
While testing on multiple datasets I've observed that if `n_iter` is low, the tuned model has lower performance for both train and test sets compared to the model using default parameters. Is there a way to ensure that the default parameters of a model are tried out as the first trial and it starts optimizing from there? If there's no improvement, there's no need to run all the trials

krfricke:
This Ray PR adds `points_to_evaluate` for all search algorithms (except random search): https://github.com/ray-project/ray/pull/12790

This PR here adds tests for this in tune-sklearn. Points that should be evaluated first can be passed via the `search_kwargs` argument. E.g.:

```
        points = [{
            "alpha": 0.4,
            "epsilon": 0.01,
            "penalty": "elasticnet"
        }, {
            "alpha": 0.3,
            "epsilon": 0.02,
            "penalty": "l1"
        }]

        tune_search = TuneSearchCV(
            self.clf,
            self.parameter_grid,
            search_optimization=search_method,
            cv=2,
            n_trials=3,
            n_jobs=1,
            refit=True,
            points_to_evaluate=points)
```

Closes #73

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install -U pytest
         python -m pip install codecov
-        python -m pip install -U https://s3-us-west-2.amazonaws.com/ray-wheels/latest/ray-1.1.0.dev0-cp37-cp37m-manylinux2014_x86_64.whl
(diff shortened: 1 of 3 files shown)


Yard1:
Looks like something breaks with Hyperopt

krfricke:
Might fix itself after https://github.com/ray-project/tune-sklearn/pull/160 gets merged.

Yard1 (approved):

PR #162, opened 2020-12-17
Bump version for 0.2.0 release

richardliaw:
Bump to 0.2.0!


setup.cfg
```diff
@@ -0,0 +1,2 @@
+[metadata]
+description-file = README.md
```
richardliaw, line 2:
```suggestion
description-file = README.md

```

setup.py
```diff
@@ -1,12 +1,23 @@
-from setuptools import setup
+import io
+import os
+from setuptools import setup, find_packages
+from tune_sklearn import __version__
+
+ROOT_DIR = os.path.dirname(__file__)
+
+VERSION = os.environ.get("TSK_RELEASE_VERSION", __version__)
```
krfricke, line 8: Why are we checking for this environment variable?

krfricke (approved):
I just have a question regarding the env variable.

richardliaw:
So that it is easy to do release candidates :)

setup.py
richardliaw: makes it easier to do release candidates / dryruns :)

PR #168, opened 2020-12-19
Fix pipeline_auto_early_stop

Yard1:
An oversight in `_PipelineTrainable` caused the xgboost/lightgbm/catboost methods to try and get boosters from the Pipeline object instead of the actual estimator, causing an exception. This PR fixes that and adds unit tests to check for this in the future.


richardliaw (approved):

richardliaw:
Awesome!!

PR #171, opened 2020-12-22
sklearn 0.24 compatibility

Yard1:
This changes the name of `self.num_samples` attribute to `self.n_trials` in `TuneSearchCV` in order to fix an exception with scikit-learn>=0.24.0. Backwards compatible with previous scikit-learn releases. API compatibility is *technically* broken, but I don't think anyone would access `num_samples` from a `TuneSearchCV` object directly.

The CI fails due to an issue with CatBoost and not tune-sklearn (looks like they were taken by surprise by the sklearn update too!).


Yard1:
Looks like the catboost example is now causing CI to fail

PR #176, opened 2021-01-14
Update code snippets in README to showcase ray.tune SearchSpaces

mkretsch327:
Small PR to update the example code snippets in the README to illustrate how users can leverage SearchSpaces to define param distributions in `TuneSearchCV`.
- Update definition of parameter distributions to illustrate new
ways of specifying parameter distributions with SearchSpaces.
- Illustrates ways to sample over discrete distributions in a couple of ways

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ notifications:
 
 install:
   - pip3 install -U -q pip
-  - pip3 install -U -q scikit-optimize==0.8.1 hyperopt==0.2.5 hpbandster==0.7.4 ConfigSpace==0.4.10 scipy dataclasses optuna==2.3.0
+  - pip3 install -U -q scikit-optimize hyperopt hpbandster ConfigSpace scipy dataclasses optuna==2.3.0
   - pip3 install -q -r requirements-test.txt
   - pip3 install --upgrade -q keras
   - pip3 install -U -q scikit-learn pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ notifications:
 
 install:
   - pip3 install -U -q pip
-  - pip3 install -U -q scikit-optimize hyperopt hpbandster ConfigSpace scipy dataclasses optuna
+  - pip3 install -U -q scikit-optimize==0.8.1 hyperopt==0.2.5 hpbandster==0.7.4 ConfigSpace==0.4.10 scipy dataclasses optuna==2.3.0
   - pip3 install -q -r requirements-test.txt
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -159,6 +161,17 @@ bohb_tune_search = TuneSearchCV(SGDClassifier(),
...
 bohb_tune_search.fit(X_train, y_train)
 
+# Define the `param_dists using the SearchSpace API
+# This allows the specification of sampling from discrete and 
+# categorical distributions (below for the `learning_rate` scheduler parameter)
+param_dists = {
+    'loss': tune.choice(['squared_hinge', 'hinge']),
+    'alpha': tune.loguniform(1e-4, 1e-1),
+    'epsilon': tune.uniform(1e-2, 1e-1),
+    'learning_rate': tune.sample_from(lambda _: np.random.choice(['constant', 'optimal'], p=[0.2, 0.8])),
```
richardliaw, line 171: hmm does the `sample_from` actually work with `hyperopt`?

.travis.yml
```diff
@@ -50,7 +50,7 @@ notifications:
 
 install:
   - pip3 install -U -q pip
-  - pip3 install -U -q scikit-optimize hyperopt hpbandster ConfigSpace scipy dataclasses optuna
+  - pip3 install -U -q scikit-optimize==0.8.1 hyperopt==0.2.5 hpbandster==0.7.4 ConfigSpace==0.4.10 scipy dataclasses optuna==2.3.0
```
richardliaw, line 53: hmm, could we keep these unpinned for tests? I think we want to be as close to latest releases as possible!

richardliaw (commented):
this looks great, thanks a bunch @mkretsch327! Just left a couple questions.

README.md
mkretsch327: Indeed it doesn't! I just pushed a change removing it. I must have polluted my notebook when I was testing this out.

.travis.yml
mkretsch327: I removed the versions for everything but optuna. It looks like they pushed a v 2.4.0 really recently which is causing some changes in their api and breaks upstream. From what I can tell this was causing some of the tests to break fail. Not sure what your philosophy is on a flakiness issue like that is, but will adjust accordingly.

I did notice this: https://github.com/ray-project/ray/blob/master/python/requirements/requirements_tune.in , which lists 2.3 for optuna at the moment.

richardliaw (approved):
Thanks a bunch for the contribution!

PR #178, opened 2021-01-27
Fix optuna CI by pinning + move to latest ray master

richardliaw:
cc @krfricke


krfricke (approved):
By the way, should we consider using dependabot here, too?

Yard1:
Is there some sort of a version check for Optuna in Tune? If it won't work with the current version it'd be best to let users know

richardliaw:
ah @Yard1 the master version of Ray will work with all versions of Optuna! (including 2.4)

PR #180, opened 2021-02-03
Fix random-search

richardliaw:
Random search with tune search spaces currently fails to produce multiple trials.


Yard1 (approved):
All it needed was that one line? Nice, hah

Yard1:
BTW. The issue with ConfigSpace is something I have experienced locally in a different project as well. Not sure what needs to be done to fix it.

Issue #186, opened 2021-03-02
[bug] TuneSearchCV with hyperopt raise an error with sklearn.RandomForestClassifier


mariesosa:
# Describe the bug

When a `TuneSearchCV` is performed with an unfitted `sklearn.RandomForestClassifier` with `search_optimization="hyperopt"` it raise the following error: `AttributeError: 'RandomForestClassifier' object has no attribute 'estimators_'`. 

# Steps/Code to Reproduce

```python
from tune_sklearn import TuneSearchCV
from ray import tune
from sklearn.ensemble import RandomForestClassifier
from sklearn.datasets import make_blobs

# Create a test dataset
X, y = make_blobs(n_samples=1000, centers=2, n_features=5)
# Instantiate a RandomForestClassifier
clf = RandomForestClassifier()
# Define a paramater to optimize
params = {'max_depth': tune.randint(5, 100)}

# Perform hyperparameter optimization
tune_search = TuneSearchCV(
    clf, params, cv=5, scoring='roc_auc', random_state=0,
    search_optimization="hyperopt"
)
tune_search = tune_search.fit(X, y)
```

# Expected Results
No error is thrown.

# Actual Results

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-ecf3059b07c2> in <module>
      4 )
      5 
----> 6 tune_search.fit(X, y)

~/virtualenvs/tf-gpu-2.2/lib/python3.6/site-packages/tune_sklearn/tune_basesearch.py in fit(self, X, y, groups, **fit_params)
    662                                     "To show process output, set verbose=2.")
    663 
--> 664             result = self._fit(X, y, groups, **fit_params)
    665 
    666             if not ray_init and ray.is_initialized():

~/virtualenvs/tf-gpu-2.2/lib/python3.6/site-packages/tune_sklearn/tune_basesearch.py in _fit(self, X, y, groups, **fit_params)
    563 
    564         self._fill_config_hyperparam(config)
--> 565         analysis = self._tune_run(config, resources_per_trial)
    566 
    567         self.cv_results_ = self._format_results(self.n_splits, analysis)

~/virtualenvs/tf-gpu-2.2/lib/python3.6/site-packages/tune_sklearn/tune_search.py in _tune_run(self, config, resources_per_trial)
    713                 "ignore", message="fail_fast='raise' "
    714                 "detected.")
--> 715             analysis = tune.run(trainable, **run_args)
[...]

PR #196, opened 2021-03-12
Pass estimators via object store IDs

krfricke:
Currently estimator objects are passed via the `config` parameter (as is the data). However, in some cases this can lead to unexpected errors when creating the search space with Hyperopt.

By passing the estimators via the object store we circumvent this problem.

After the next release we might want to use `tune.with_parameters()` to avoid passing anything other than the search space via the config parameter, see https://github.com/ray-project/ray/pull/14532

Closes #https://github.com/ray-project/tune-sklearn/issues/186

### DIFF
--- a/tune_sklearn/_trainable.py
+++ b/tune_sklearn/_trainable.py
@@ -1,6 +1,5 @@
 """ Helper class to train models using Ray backend
 """
-from sklearn.base import clone
 from sklearn.model_selection import cross_validate
 from sklearn.utils.metaestimators import _safe_split
 import numpy as np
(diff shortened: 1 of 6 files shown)


tune_sklearn/_trainable.py
```diff
@@ -43,7 +42,8 @@ def _setup(self, config):
                 stopping if it is set to true.
 
         """
-        self.estimator_list = clone(config.pop("estimator_list"))
+        estimator_ids = list(config.pop("estimator_ids"))
+        self.estimator_list = ray.get(estimator_ids)
```
krfricke, line 46: This implicitly clones the estimators

Yard1 (approved):

PR #198, opened 2021-03-23
Allow any searcher, fixes

Yard1:
This PR:
* Allows the user to pass any Searcher subclass to TuneSearchCV, making it possible to use Searchers other than predefined
* Fixes an issue with multimetric scoring where values for `refit` parameter were not being enforced as intended
* Moves `metric` and `mode` from Searchers/Schedulers to `tune.run` where possible

There are no API breaking changes and CI should pass.

The reason why a Searcher subclass is passed instead of an object is that I can't see a good way of handling various parameters inside it. For example, someone may pass an object with `_metric` and `_mode` already set which would cause issues if those keys were passed to `tune.run` - but it's also not possible to just check for those attributes as there's no requirement of them being there. Forcing the user to pass a class instead (while still allowing them to instantiate it with custom kwargs) seemed like the best solution.

### DIFF
--- a/examples/custom_searcher_example.py
+++ b/examples/custom_searcher_example.py
@@ -1,0 +1,42 @@
+"""
+An example using a Searcher (HEBOSearch) not defined
+directly in TuneSearchCV.
+"""
+
+from tune_sklearn import TuneSearchCV
+from ray import tune
+from ray.tune.suggest.hebo import HEBOSearch
+from sklearn.ensemble import RandomForestClassifier
+from sklearn import datasets
+from sklearn.model_selection import train_test_split
+import numpy as np
+
+digits = datasets.load_digits()
+x = digits.data
+y = digits.target
+x_train, x_test, y_train, y_test = train_test_split(x, y, test_size=.2)
+
+clf = RandomForestClassifier()
+param_distributions = {
+    "n_estimators": tune.randint(20, 80),
+    "max_depth": tune.randint(2, 10)
+}
+
+# If a Searcher is initialized without specifying search space,
+# it will use the space passed to TuneSearchCV
+# (in this example: param_distributions)
+searcher = HEBOSearch()
+
+# It is also possible to use user-defined Searchers, as long as
+# they inherit from ray.tune.suggest.Searcher and have the following
+# attributes: _space, _metric, _mode
+
+tune_search = TuneSearchCV(
+    clf, param_distributions, n_trials=3, search_optimization=searcher
+)
+
+tune_search.fit(x_train, y_train)
+
+pred = tune_search.predict(x_test)
+accuracy = np.count_nonzero(np.array(pred) == np.array(y_test)) / len(pred)
+print(accuracy)
--- a/examples/custom_searcher_example.py
+++ b/examples/custom_searcher_example.py
@@ -32,8 +32,7 @@ searcher = HEBOSearch()
 # attributes: _space, _metric, _mode
 
 tune_search = TuneSearchCV(
-    clf, param_distributions, n_trials=3, search_optimization=searcher
-)
+    clf, param_distributions, n_trials=3, search_optimization=searcher)
 
 tune_search.fit(x_train, y_train)
 
--- a/tune_sklearn/tune_gridsearch.py
+++ b/tune_sklearn/tune_gridsearch.py
@@ -139,6 +139,8 @@ class TuneGridSearchCV(TuneBaseSearchCV):
         time_budget_s (int|float|datetime.timedelta): Global time budget in
             seconds after which all trials are stopped. Can also be a
             ``datetime.timedelta`` object.
+        mode (str): One of {min, max}. Determines whether objective is
+            minimizing or maximizing the metric attribute. Defaults to "max".
     """
 
     def __init__(self,
@@ -159,7 +161,8 @@ class TuneGridSearchCV(TuneBaseSearchCV):
                  pipeline_auto_early_stop=True,
                  stopper=None,
                  time_budget_s=None,
-                 sk_n_jobs=None):
+                 sk_n_jobs=None,
+                 mode=None):
         if sk_n_jobs is not None:
             raise ValueError(
                 "Tune-sklearn no longer supports nested parallelism "
@@ -180,7 +183,8 @@ class TuneGridSearchCV(TuneBaseSearchCV):
             loggers=loggers,
             pipeline_auto_early_stop=pipeline_auto_early_stop,
             stopper=stopper,
-            time_budget_s=time_budget_s)
+            time_budget_s=time_budget_s,
+            mode=mode)
 
         check_error_warm_start(self.early_stop_type, param_grid, estimator)
 
@@ -261,7 +265,7 @@ class TuneGridSearchCV(TuneBaseSearchCV):
             loggers=self.loggers,
             time_budget_s=self.time_budget_s,
             metric=self._metric_name,
-            mode="max")
+            mode=self.mode)
 
         if isinstance(self.param_grid, list):
             run_args.update(
--- a/tune_sklearn/tune_gridsearch.py
+++ b/tune_sklearn/tune_gridsearch.py
@@ -259,7 +259,9 @@ class TuneGridSearchCV(TuneBaseSearchCV):
             resources_per_trial=resources_per_trial,
             local_dir=os.path.expanduser(self.local_dir),
             loggers=self.loggers,
-            time_budget_s=self.time_budget_s)
+            time_budget_s=self.time_budget_s,
+            metric=self._metric_name,
+            mode="max")
 
         if isinstance(self.param_grid, list):
             run_args.update(
--- a/tune_sklearn/tune_search.py
+++ b/tune_sklearn/tune_search.py
@@ -37,9 +37,6 @@ available_optimizations = {
     OptunaSearch: "optuna",
 }
 
-# sklearn always maximizes
-DEFAULT_MODE = "max"
-
 
 def _check_distribution(dist, search_optimization):
     # Tune Domain is always good
@@ -298,6 +295,8 @@ class TuneSearchCV(TuneBaseSearchCV):
             seconds after which all trials are stopped. Can also be a
             ``datetime.timedelta`` object. The stopping condition is checked
             after receiving a result, i.e. after each training iteration.
+        mode (str): One of {min, max}. Determines whether objective is
+            minimizing or maximizing the metric attribute. Defaults to "max".
         **search_kwargs (Any):
             Additional arguments to pass to the SearchAlgorithms (tune.suggest)
             objects.
@@ -326,6 +325,7 @@ class TuneSearchCV(TuneBaseSearchCV):
                  stopper=None,
                  time_budget_s=None,
                  sk_n_jobs=None,
+                 mode=None,
                  **search_kwargs):
         if sk_n_jobs is not None:
             raise ValueError(
@@ -402,7 +402,8 @@ class TuneSearchCV(TuneBaseSearchCV):
             loggers=loggers,
             pipeline_auto_early_stop=pipeline_auto_early_stop,
             stopper=stopper,
-            time_budget_s=time_budget_s)
+            time_budget_s=time_budget_s,
+            mode=mode)
 
         check_error_warm_start(self.early_stop_type, param_distributions,
                                estimator)
@@ -666,7 +667,7 @@ class TuneSearchCV(TuneBaseSearchCV):
             loggers=self.loggers,
             time_budget_s=self.time_budget_s,
             metric=self._metric_name,
-            mode=DEFAULT_MODE)
+            mode=self.mode)
 
         if self.search_optimization == "random":
             if isinstance(self.param_distributions, list):
@@ -687,13 +688,13 @@ class TuneSearchCV(TuneBaseSearchCV):
                         f"('{self.search_optimization._metric}') "
                         "must match the metric set in TuneSearchCV"
                         f" ('{self._metric_name}')")
-                if self.search_optimization._mode != DEFAULT_MODE:
+                if self.search_optimization._mode != self.mode:
                     raise ValueError(
                         "If a Searcher instance has been initialized with a "
                         "space, its mode "
                         f"('{self.search_optimization._mode}') "
                         "must match the mode set in TuneSearchCV"
-                        f" ('{DEFAULT_MODE}')")
+                        f" ('{self.mode}')")
             elif self._is_param_distributions_all_tune_domains():
                 run_args["config"].update(self.param_distributions)
                 override_search_space = False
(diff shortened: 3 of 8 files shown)


tune_sklearn/tune_search.py
```diff
@@ -656,17 +673,29 @@ def _tune_run(self, config, resources_per_trial):
...
+                            run_args["scheduler"], "_mode"):
+                        run_args["scheduler"]._metric = search_kwargs["metric"]
+                        run_args["scheduler"]._mode = search_kwargs["mode"]
+                    else:
+                        warnings.warn(
+                            "Could not set `_metric` and `_mode` attributes "
+                            f"on Scheduler {run_args['scheduler']}. "
+                            "This may cause an exception later! "
+                            "Ensure your Scheduler initializes with those "
+                            "attributes.", UserWarning)
```
richardliaw, line 690: is there some way of setting this, like with `set_search_properties`?

tune_sklearn/tune_search.py
```diff
@@ -12,7 +12,13 @@
...
 from ray import tune
 from ray.tune.sample import Domain
-from ray.tune.suggest import ConcurrencyLimiter, BasicVariantGenerator
+from ray.tune.suggest import (ConcurrencyLimiter, BasicVariantGenerator,
+                              Searcher)
+from ray.tune.suggest.bohb import TuneBOHB
+from ray.tune.schedulers import HyperBandForBOHB
+from ray.tune.suggest.skopt import SkOptSearch
+from ray.tune.suggest.hyperopt import HyperOptSearch
+from ray.tune.suggest.optuna import OptunaSearch, param
```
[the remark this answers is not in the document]
Yard1: All of those are included in Ray Tune, they just won't work without their own dependencies - however, imports by themselves don't need anything extra to be installed.

tune_sklearn/tune_search.py
```diff
@@ -306,17 +329,16 @@ def __init__(self,
...
+        if isinstance(search_optimization, str):
+            search_optimization = search_optimization.lower()
+
+        if (search_optimization not in set(available_optimizations.values())
+            ) and (not isinstance(search_optimization, type)
+                   or not issubclass(search_optimization, Searcher)):
+            raise ValueError(
+                "Search optimization must be one of "
+                f"{', '.join(list(available_optimizations.values()))} "
+                "or a ray.tune.suggest.Searcher class.")
```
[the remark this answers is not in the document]
Yard1: I couldn't figure out a good way to pass an already initialized object, as I have detailed in description - tl;dr there is no enforcement of attributes on Searcher subclassses, so while using objects would work with the Searchers present in Tune, it may not work with user-defined ones

tune_sklearn/tune_search.py
Yard1: We can't use `set_search_properties` as that would cause issues when it's passed to `tune.run`, which calls it itself - and it expects an object which hadn't had it called before

tune_sklearn/tune_search.py
```diff
@@ -388,13 +409,19 @@ def __init__(self,
...
+                search_optimization, type):
             if search_kwargs:
-                raise ValueError("Random search does not support "
+                raise ValueError(f"{search_optimization} does not support "
                                  f"extra args: {search_kwargs}")
         self.search_optimization = search_optimization
         self.search_kwargs = search_kwargs
 
+    @property
+    def _str_search_optimization(self):
```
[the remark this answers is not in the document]
Yard1: Done

tune_sklearn/tune_search.py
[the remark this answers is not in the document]
krfricke: Hm, so Tune uses `set_search_properties()` to set the search space, metric, and mode of a search algorithm. The `Searcher` base class also has a `_metric` and `_mode` attribute. Can't we assume that custom searchers always implement that interface?
Custom searchers were not possible to use before with tune-sklearn, right? Thus, now would be a good time to define and enforce properties of these custom searchers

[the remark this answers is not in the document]
Yard1: I'd think that would need to be enforced in Ray Tune itself, but we can check it here for now. In that case, I'll make it so that an object is passed instead of a class

richardliaw:
Sorry for the slow reply!

> The reason why a Searcher subclass is passed instead of an object is that I can't see a good way of handling various parameters inside it. For example, someone may pass an object with _metric and _mode already set which would cause issues if those keys were passed to tune.run 

If I had a custom Searcher, how would I pass it in to TuneSearchCV with custom arguments?

I actually think we could just check if `object.metric` / `object.mode` is set, and fail if they are?

Yard1:
@richardliaw You'd pass custom Searcher args as kwargs in TuneSearchCV - however now that we've talked it through it will probably be much cleaner to pass Searcher objects instead and enforce those attributes on them. I'll change it after holidays here :)

Yard1:
I had a look at it and using objects is not as straightforward.

The issue is that if a user passes an algo specific search space, which can't be passed directly to `tune.run` (like the `ConfigSpace` object for BOHB) then the handling for that is usually present in the Searcher's `__init__`. It also sets things like `metric_op` and such, which can vary from algo to algo. It is thus not possible to simply change the attributes and pass the changed object to `tune.run`.

Therefore, if we want to allow users to pass Searcher instances, instead of classes as I first proposed, we would need to force the user to use only Tune search spaces with them.

What are your thoughts on solving this? I agree that passing classes instead of instances is not intuitive but I am not sure what'd be a solution that would not have us prevent allowing algo specific search space objects.

richardliaw:
@Yard1 could you help me understand the following questions? Under the object searcher proposal:

1. Will users be able to provide a string argument for the searcher?
2. If using a custom searcher object, why would the user not be able to just pass the custom config space via the searcher object?
3. What are the breaking changes made to user experience by using the object searcher proposal?

Yard1:
@richardliaw
1. That is unchanged in either case (no API breaking changes)
2. Yes, but they will also need to initialize the object with metric and mode alongside the space. All those attributes would be detached from the space, metric and mode in TuneSearchCV, as it will be not possible to set them in a way that will cover all possible cases. There are a lot of algo specific interactions and derived attributes that are set on initialization - for example, many Searchers set the `metric_op` attribute during initialization depending on the `_mode` attribute. We could technically make cases for all Searcher classes in Ray Tune but it will be a nightmare to maintain and it will not account for the possibility of a user making their own Searcher. 
3. As described in 2, it would completely detach the parameters set in TuneSearchCV with what's set on the Searcher object. I can imagine a situation where someone would forget to initialize  a new Searcher instance with a different metric, changing it only in TuneSearchCV. For the same reason, ensuring consistency just by raising exceptions in case of a mismatch will not be robust either (though more doable than setting them, as derived attributes will not need to be accounted for).

richardliaw:
One use case I'd like to support is the search algorithm restoration. This allows people to use learnings from past optimization jobs to speed up new ones.

This would require access to the search algorithm object (Searcher.restore()). 

Ideally, we would have the following:

```
# 1. Base usage if using a custom searcher
searcher = Searcher()
TuneSearchCV(searcher, config=config, metric=X, mode=Y)

# 2. RAISES ERROR if metrics are set in searcher
searcher = Searcher(metric=X, mode=Y)
TuneSearchCV(searcher, metric=X, mode=Y)

# 3. custom search space, OK
searcher = Searcher(config=config)  # is this possible?
TuneSearchCV(searcher, metric=X, mode=Y)

# 4. RAISES error if both config is set.
searcher = Searcher(config=config) 
TuneSearchCV(config=config, searcher=searcher, metric=X, mode=Y)

# 5. Restore from previous run
searcher = Searcher()
searcher.restore("path")
TuneSearchCV(config=config, searcher=searcher, metric=X, mode=Y)
```

My understanding is that 3 is the controversial one? Though actually a quick skim of code tells me that you can actually pass space without setting metric/mode for many search algs.

Yard1:
So there are two ways of configuring searchers. Either through `set_search_properties` called by `tune.run` or by specifying the necessary parameters in `__init__`. The issue is that for all the Tune searchers, if the instance is initialized with the space, then the `set_search_properties` will fail and `tune.run` will raise an exception. You can see this behavior in eg. the Optuna Searcher: 

```python
class OptunaSearch(Searcher):
    def __init__(self,
                 space: Optional[Union[Dict, List[Tuple]]] = None,
                 metric: Optional[str] = None,
                 mode: Optional[str] = None,
                 points_to_evaluate: Optional[List[Dict]] = None,
                 sampler: Optional[BaseSampler] = None):
        # ...

        if isinstance(space, dict) and space:
            resolved_vars, domain_vars, grid_vars = parse_spec_vars(space)
            if domain_vars or grid_vars:
                logger.warning(
                    UNRESOLVED_SEARCH_SPACE.format(
                        par="space", cls=type(self)))
                space = self.convert_search_space(space)

        self._space = space
        # ...
        if self._space:
            self._setup_study(mode)

    def set_search_properties(self, metric: Optional[str], mode: Optional[str],
                              config: Dict) -> bool:
        if self._space:
            return False
        space = self.convert_search_space(config)
        self._space = space
        if metric:
            self._metric = metric
        if mode:
            self._mode = mode

        self._setup_study(mode)
        return True
```

If the space is set in `__init__`, then it is not possible to set metric and mode through `set_search_properties`. All the other Searchers have the same behavior. Therefore, there are only two ways:

1. If space, metric and mode are to be set through `set_search_properties` called by `tune.run`, then the Searcher has to be initialized *without* specifying the space, so that `set_search_properties` sets all three
2. Alternatively, all three attributes must be set in `__init__`.

It is not possible to eg. set just the space in `__init__` and then metric and mode by `tune.run` as that will raise an excception. On the other hand, `tune.run` only supports dicts as search spaces so it is not possible to pass algo-specific search space objects to it, leaving only the second choice available.

I see two solutions:
1. Force everything to go through `tune.run` - will be the simplest, most straightforward and easiest to maintain option but it will make it impossible to eg. pass a ConfigSpace object directly to BOHB, forcing the user to recreate it as a dict of Tune search spaces
2. If the searcher has been initialized with a space already, then raise exceptions if metrics and mode in TuneSearchCV are different - this is not 100% robust, but should work fine for most cases.

I hope this clears it up, sorry for the confusion. The gist is that it's only possible to either set space, metric and mode all at once in `__init__` of a Searcher, or none at all.

richardliaw:
Hmm ok thanks for the clarification!

What about we go with: 

> If the searcher has been initialized with a space already, then raise exceptions if metrics and mode in TuneSearchCV are different - this is not 100% robust, but should work fine for most cases.

But also make it so that we don't actually need to specify it twice (i.e., you can leave the one in TuneSearchCV empty?)

Yard1:
So the properties defined in the Searcher should overwrite the ones in TuneSearchCV? The ones in TuneSearchCV also used for early stopping (Scheluders).

richardliaw:
> So the properties defined in the Searcher should overwrite the ones in TuneSearchCV? The ones in TuneSearchCV also used for early stopping (Scheluders).

Ah, no. Ok to make it simple, how about you just make sure everyone has to have the same for now, and we can think about overriding later?

TuneSearchCV mode/metric has to be set, and Scheduler/Searchers will also have to be set if passed in.

Yard1:
Sure, let me get on that

Yard1:
It's now using instances while enforcing consistency between TuneSearchCV and the Searcher.

examples/custom_searcher_example.py
```diff
@@ -0,0 +1,41 @@
...
+# it will use the space passed to TuneSearchCV
+# (in this example: param_distributions)
+searcher = HEBOSearch()
+
+# It is also possible to use user-defined Searchers, as long as
+# they inherit from ray.tune.suggest.Searcher and have the following
+# attributes: _space, _metric, _mode
+
+tune_search = TuneSearchCV(
+    clf, param_distributions, n_trials=3, search_optimization=searcher)
```
richardliaw, line 35: so in this case you don't need to specify metric/mode (because it's automatic)?

tune_sklearn/tune_gridsearch.py
```diff
@@ -259,7 +259,9 @@ def _tune_run(self, config, resources_per_trial):
             resources_per_trial=resources_per_trial,
             local_dir=os.path.expanduser(self.local_dir),
             loggers=self.loggers,
-            time_budget_s=self.time_budget_s)
+            time_budget_s=self.time_budget_s,
+            metric=self._metric_name,
+            mode="max")
```
richardliaw, line 264: hm is it possible to make 'mode' configurable?

richardliaw: I guess it is not a regression?

tune_sklearn/tune_search.py
```diff
@@ -22,12 +28,27 @@
...
+    BasicVariantGenerator: "random",
+    RandomListSearcher: "random",
+    SkOptSearch: "bayesian",  # scikit-optimize/SkOpt
+    TuneBOHB: "bohb",
+    HyperOptSearch: "hyperopt",
+    OptunaSearch: "optuna",
+}
+
+# sklearn always maximizes
+DEFAULT_MODE = "max"
```
richardliaw, line 41: move to utils?

tune_sklearn/tune_search.py
```diff
@@ -651,22 +677,51 @@ def _tune_run(self, config, resources_per_trial):
...
             search_kwargs = self.search_kwargs.copy()
-            search_kwargs.update(metric=self._metric_name, mode="max")
+            if override_search_space:
+                search_kwargs["metric"] = run_args.pop("metric")
+                search_kwargs["mode"] = run_args.pop("mode")
+                if run_args["scheduler"]:
+                    if hasattr(run_args["scheduler"], "_metric") and hasattr(
+                            run_args["scheduler"], "_mode"):
+                        run_args["scheduler"]._metric = search_kwargs["metric"]
+                        run_args["scheduler"]._mode = search_kwargs["mode"]
```
richardliaw, line 709: let's raise an issue on ray-project/ray for this

richardliaw (approved):
Nice this looks great!

richardliaw:
@Yard1 could you look into the test and see if this is related?

tune_sklearn/tune_gridsearch.py
richardliaw: (feel free to resolve)

Yard1:
@richardliaw CI is failing due to Optuna - I believe the version is wrong, I'll check - but it shouldn't be related to this.

tune_sklearn/tune_gridsearch.py
Yard1: We could make it configurable, shouldn't be an issue

examples/custom_searcher_example.py
Yard1: If metric, mode and space are unspecified in the Searcher, then the ones defined in TuneSearchCV will be used (and it has its default ones)

Yard1:
@richardliaw Mode is now configurable and release CI passes (it was installing a version of Optuna that was too old)

richardliaw:
Great work @Yard1 !

Issue #211, opened 2021-06-04
[feature request] Support `resume` (and potentially other `tune.run` args)


Yard1:
From: https://discuss.ray.io/t/tunesearchcv-resume/2414

> Is there a way to resume TuneSearchCV like with tune.run? I am using named log folders, but it seems like the resume parameter is missing–Is it possible to add? or would you recommend I try to find a workaround, like using the vanilla tune.run?

I think we could just have a `tune_run_kwargs` argument somewhere, probably in `fit` that would allow the user to pass any parameter they want to `tune.run`, including `resume`.

PR #212, opened 2021-06-07
Allow user to pass own `tune.run` params in `fit`

Yard1:
Will fix https://github.com/ray-project/tune-sklearn/issues/211

### DIFF
--- a/tune_sklearn/tune_basesearch.py
+++ b/tune_sklearn/tune_basesearch.py
@@ -747,15 +747,15 @@ class TuneBaseSearchCV(BaseSearchCV):
                 configuration for `tune.run`.
             resources_per_trial (:obj:`dict` of int): dictionary specifying the
                 number of cpu's and gpu's to use to train the model.
-            tune_params (dict, optional): User defined parameters passed to
-                ``tune.run``.
+            tune_params (dict): User defined parameters passed to
+                ``tune.run``. Parameters inside `tune_params` override
+                preset parameters.
 
         """
         raise NotImplementedError("Define in child class")
 
     def _override_run_args_with_tune_params(self, run_args, tune_params):
-        """Helper method to override tune.run run arguments with user
-        supplied dict"""
+        """Helper to override tune.run run arguments with user supplied dict"""
 
         if tune_params:
             user_overrides = {k for k in tune_params if k in run_args}
@@ -766,7 +766,6 @@ class TuneBaseSearchCV(BaseSearchCV):
                     ". This may cause unexpected issues! If you experience "
                     "issues, please try removing those parameters from "
                     "tune_params.")
-            warnings.warn("Using user supplied tune_params.")
             run_args = {**run_args, **tune_params}
         return run_args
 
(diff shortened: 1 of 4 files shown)


tune_sklearn/tune_basesearch.py
```diff
@@ -735,18 +739,37 @@ def _fill_config_hyperparam(self, config):
...
+        if tune_params:
+            user_overrides = {k for k in tune_params if k in run_args}
+            if user_overrides:
+                warnings.warn(
+                    "The following preset tune.run parameters will "
+                    f"be overriden by tune_params: {', '.join(user_overrides)}"
+                    ". This may cause unexpected issues! If you experience "
+                    "issues, please try removing those parameters from "
+                    "tune_params.")
+            warnings.warn("Using user supplied tune_params.")
```
richardliaw, line 769: hmm, if a user passes this in, they wouldn't expect a warning here right?

tune_sklearn/tune_basesearch.py
```diff
@@ -735,18 +739,37 @@ def _fill_config_hyperparam(self, config):
...
+    def _override_run_args_with_tune_params(self, run_args, tune_params):
+        """Helper method to override tune.run run arguments with user
+        supplied dict"""
+
+        if tune_params:
+            user_overrides = {k for k in tune_params if k in run_args}
+            if user_overrides:
+                warnings.warn(
+                    "The following preset tune.run parameters will "
+                    f"be overriden by tune_params: {', '.join(user_overrides)}"
```
richardliaw, line 765: this is a great error message!

tune_sklearn/tune_search.py
```diff
@@ -626,6 +626,8 @@ def _tune_run(self, config, resources_per_trial):
             resources_per_trial (dict): Resources to use per trial within Ray.
                 Accepted keys are `cpu`, `gpu` and custom resources, and values
                 are integers specifying the number of each resource to use.
+            tune_params (dict, optional): User defined parameters passed to
```
richardliaw, line 629: I think to get this to properly render on readthedocs, you can just leave this as (dict)

richardliaw: Can you add a comment here also about the overriding behavior?

tune_sklearn/tune_basesearch.py
```diff
@@ -735,18 +739,37 @@ def _fill_config_hyperparam(self, config):
...
                 number of cpu's and gpu's to use to train the model.
+            tune_params (dict, optional): User defined parameters passed to
+                ``tune.run``.
 
         """
         raise NotImplementedError("Define in child class")
 
+    def _override_run_args_with_tune_params(self, run_args, tune_params):
+        """Helper method to override tune.run run arguments with user
+        supplied dict"""
```
richardliaw, line 758: nit (not important here but for future ref), the first line of the docstring is a summary line - and should not have a linebreak in the middle. this is important when building docs

richardliaw:
Sorry I missed this! left some simple comments. Please assign me when you get around to addressing the issues!

tune_sklearn/tune_basesearch.py
Yard1: This probably should be an logging info message instead of a warning, now that I think about it

Yard1:
Thanks for letting me know about the docs!

richardliaw (approved):

PR #224, opened 2021-09-30
Save analysis object

Yard1:


requirements-test.txt
```diff
@@ -1,5 +1,5 @@
 numpy>=1.16
-scikit-learn
+scikit-learn<1.0
```
xwjiang2010, line 2: why do we need to pin the version?

xwjiang2010 (approved):

requirements-test.txt
Yard1: Otherwise tests fail due to skopt not being up to date with sklearn 1.0

Issue #221, opened 2021-09-23
ValueError: Tune-sklearn no longer supports nested parallelism with new versions of joblib/sklearn. Don't set 'sk_n_jobs'.


RNarayan73:
Hello,

I'm trying to use TuneSearchCV in a pipeline nested within the cross_validate function. As per sklearn's [recommendation](https://scikit-learn.org/stable/auto_examples/model_selection/plot_nested_cross_validation_iris.html#sphx-glr-auto-examples-model-selection-plot-nested-cross-validation-iris-py) this will enable a more robust evaluation of the generalisation error. Note, this isn't a cross validation for hyperparameter optimisation within TuneSearchCV, but of the overall pipeline which includes TuneSearchCV as the last step.
![image](https://user-images.githubusercontent.com/82275999/134586169-99f60731-161a-447f-aa3e-e1878cad6c49.png)

I get the above error when I execute cross_validate even though sk_n_jobs is left as the default value (None). 
Furthermore, when I explicitly set sk_n_jobs to 1 (in the hope of avoiding parallelism) I get the same error earlier, when I try to instantiate TuneSearchCV.

How can I avoid this and run cross_validation? I have managed to test models this way using GridSearchCV and the SKOpt, Optuna optimisers with their SKLearn wrapper, but am stuck when trying it with TuneSearchCV.

Thanks for your help.

Narayan

Yard1:
Are you using the latest release of tune-sklearn (0.4.1)?

RNarayan73:
Yes, on 0.4.1
Here's the code:
`hpt_pipe = TuneSearchCV(pipe, 
                                param_distributions=get_params_space(algo, hpt), 
                                cv=cv_skf,   # cv object
                                scoring=score_metrics,    # dict with scoring metrics
                                refit=eval_metric, 
                                name=hpt + '-' + run, 
                                #early_stopping=False, 
                                pipeline_auto_early_stop=True,  # If True, early stopping will be performed on the last stage of the pipeline (which must support early stopping)
                                                                # If False, early stopping will be determined by ‘Pipeline.warm_start’ or ‘Pipeline.partial_fit’ capabilities, which are by default not supported by standard SKlearn
                                local_dir=TUNE_FOLDER, 
                                n_trials=ITERS, random_state=SEED, 
                                #sk_n_jobs=None,   # Tune-sklearn no longer supports nested parallelism with new versions of joblib/sklearn. Don't set 'sk_n_jobs'
[...]

PR #229, opened 2021-11-25
Fix cloning issues with early_stopping

Yard1:
Fixes sklearn cloning when `early_stopping` is modified internally.

Closes https://github.com/ray-project/tune-sklearn/issues/221


krfricke (approved):
Rubber stamping

PR #236, opened 2022-03-14
Remove `ray_context`

Yard1:
Removes `ray_context` as it's not a pattern seen in any other Ray ecosystem library.


krfricke (approved):
LGTM.

PR #237, opened 2022-03-14
Use `tune.with_parameters`

Yard1:
Uses `tune.with_parameters` instead of manually converting to object references.


krfricke (approved):
This is great. Thanks!

PR #240, opened 2022-04-18
Support sklearn.compose.TransformedTargetRegressor

aidiss:
Currently TuneSearchCV fails when provided with LGBMRegressor provided wrapped inside TransformedTargerRegressor.

For example, this block would fail

```
regressor = LGBMRegressor(**config)

regressor = TransformedTargetRegressorTwo(
    regressor=regressor, func=np.log1p, inverse_func=np.expm1
)
```
Failure happens partly because of failure of tune_search.utils.get_early_stop_type to read the type of the estimator.

Note sure if this is the correct implementation. But it should not break anything.

### DIFF
--- a/tune_sklearn/utils.py
+++ b/tune_sklearn/utils.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from typing import Dict
 
+from sklearn.compose import TransformedTargetRegressor
 from sklearn.metrics import check_scoring
 from sklearn.pipeline import Pipeline
 from tune_sklearn._detect_booster import (
@@ -94,6 +95,10 @@ def check_error_warm_start(early_stop_type, estimator_config, estimator):
 
 
 def get_early_stop_type(estimator, early_stopping):
+    # If estimator is TransformedTargetRegressor we should get the wrapped regressor.
+    if isinstance(estimator, TransformedTargetRegressor):
+        estimator = estimator.regressor
+
     if not early_stopping:
         return EarlyStopping.NO_EARLY_STOP
     can_partial_fit = check_partial_fit(estimator)


Yard1:
Hey, this is a good catch. I think we would want to make this more generic if possible. I am happy to merge this as is, though we would eventually want to support other meta-estimators (including user-defined ones). Perhaps we can allow the user to just directly specify the early stop type. Can you add a unit test for this? Thanks!

aidiss:
Added a test.
Let me know if its in the right place and done in a right way.

aidiss:
Secondly, I wonder, what could be implementation for handling estimators that are inside pipelines.
I guess we could traverse the estimators and look for the one that is supported? Or vice versa, skip the ones that are meta?
Or, maybe it could be possible to tell specifically what kind of early stopping is used by the estimator?

tune_sklearn/utils.py
```diff
@@ -101,6 +103,14 @@ def get_early_stop_type(estimator, early_stopping):
 
     if not early_stopping:
         return EarlyStopping.NO_EARLY_STOP
+
+    if check_is_pipeline(estimator):
+        for step_name, step in estimator.steps:
```
Yard1, line 108: Two things:
- we shouldn't be only checking for regressors
- this is not foolproof & goes against duck typing principle of sklearn
- we can just assume that the last step of the pipeline is an estimator. You can't really make a pipeline with multiple estimators.

Yard1:
Hey @aidiss I took a look at the logic and I unfortunately don't think this will work. We may detect the early stop type correctly, but we are unable to apply it during actual training if the target transformer is present. It would require a more through refactoring of the code. I'll put this on the backlog, unless you would be willing to work on this. We would need to support both the case with a pipeline and without it.

PR #241, opened 2022-04-21
Move from loggers to logger callbacks

krfricke:
`loggers` have been deprecated for some time - use `callbacks` instead

Closes https://github.com/ray-project/tune-sklearn/issues/239


tune_sklearn/tune_search.py
```diff
@@ -691,7 +691,8 @@ def _tune_run(self, X, y, config, resources_per_trial, tune_params=None):
             resources_per_trial=resources_per_trial,
             local_dir=self.local_dir,
             name=self.name,
-            loggers=resolve_loggers(self.loggers, self.defined_schedulers),
+            callbacks=resolve_logger_callbacks(self.loggers,
+                                               self.defined_loggers),
```
Yard1, line 695: Can we also edit the  `_override_run_args_with_tune_params` function to merge the preset logging callbacks we add with any callbacks the user would pass in `tune_params`?

Yard1 (approved):
Thanks!

PR #264, opened 2023-05-24
Import `mlflow` integration from new `air.integrations` module

justinvyu:
Ray has some unit-tests depending on this package. The `ray.tune.integration` package is getting hard-deprecated in Ray 2.6, so this needs package needs to import from the new location at `ray.air.integrations` to unblock that.

In particular, it's this user guide doctest that would fail: https://docs.ray.io/en/latest/tune/examples/tune-sklearn.html

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install -U pytest
         python -m pip install codecov
-        python -m pip install -U https://s3-us-west-2.amazonaws.com/ray-wheels/latest/ray-3.0.0.dev0-cp37-cp37m-manylinux2014_x86_64.whl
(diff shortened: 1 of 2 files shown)


Yard1 (approved):

Yard1:
Actually, can we have a try...except import here? @justinvyu

Yard1 (changes requested):
see above

justinvyu:
I think the MLFlowLoggerCallback already try catches if mlflow is not installed.

matthewdeng:
@Yard1 by try/except do you mean something like:

> try import from air --> except import from tune

justinvyu:
No need to try catch since tune_sklearn depends on ray>=2.0, and the module was already moved to air.integrations at 2.0. Relaunching the CI, then we can merge after it passes.

justinvyu:
CI for latest Ray release is failing due to a fix that did not make the branch cut

CI for master is failing due to a similar issue as this flaky example on ray: https://github.com/ray-project/ray/issues/35428

justinvyu:
Passing with master wheel here: https://github.com/ray-project/tune-sklearn/actions/runs/5426940588/jobs/9869633497

Yard1 (approved):

PR #272, opened 2023-09-06
Fix `ExperimentAnalysis` usage for Ray 2.7

justinvyu:
Ray 2.7 has some breaking API changes for `tune.ExperimentAnalysis`, which `tune-sklearn` depends on to construct the output. This PR fixes the API usage.

Closes https://github.com/ray-project/tune-sklearn/issues/273

### DIFF
--- a/tests/test_randomizedsearch.py
+++ b/tests/test_randomizedsearch.py
@@ -156,7 +156,7 @@ class RandomizedSearchTest(unittest.TestCase):
             parameter_grid,
             early_stopping=scheduler,
             max_iters=10,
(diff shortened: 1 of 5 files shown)


krfricke (commented):
The test failures don't seem related:

```
 (_Trainable pid=2230) ValueError: Unsupported set of arguments: The combination of penalty='l2' and loss='hinge' are not supported when dual=False, Parameters: penalty='l2', loss='hinge', dual=False
```

but it would be good to confirm that these changes don't break anything.

matthewdeng (approved):

PR #281, opened 2023-11-03
Bump version to `0.5.0` and update Ray Tune lower bound to `2.7.1`

justinvyu:
This reverts commit 0eb3a9f1ab2ee3f9c0927bbd3f4f57f5fc99610b.


setup.py
```diff
@@ -28,6 +28,6 @@
     long_description_content_type="text/markdown",
     url="https://github.com/ray-project/tune-sklearn",
     install_requires=[
-        "scikit-learn", "scipy", "ray[tune]>=2.0.0", "numpy>=1.16"
+        "scikit-learn", "scipy", "ray[tune]>=2.7.1", "numpy>=1.16"
```
matthewdeng, line 31: Does everything work with 2.7.1 or should we bump to 2.8 now?

matthewdeng (approved):

setup.py
justinvyu: 2.7.1 is an ok lower bound

